PR #11, opened 2026-01-02
Feature unlink symlinks during uninstall

Max13245:
Unlinks symlinks from Packit during the install by looping over all Packit directories and searching for symlinks which point to  or inside of the package that will be uninstalled.

### DIFF
--- a/src/installer/installer.rs
+++ b/src/installer/installer.rs
@@ -1,5 +1,5 @@
 use crate::{
-    cli::{self, ask_user, QuestionResponse, Spinner},
+    cli::{ask_user, QuestionResponse, Spinner},
     config::Config,
     installed_packages::InstalledPackageStorage,
     installer::{
@@ -11,7 +11,10 @@ use crate::{
     repositories::manager::RepositoryManager,
 };
 
-use std::{fs, path::Path};
+use std::{
+    fs,
+    path::{Path, PathBuf},
+};
 
 /// The installer of Packit, managing the installation of packages on the system.
 pub struct Installer<'a> {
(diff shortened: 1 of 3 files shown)


src/installer/installer.rs
```diff
@@ -234,16 +233,16 @@ impl<'a> Installer<'a> {
...
+        let directory = self.config.prefix_directory.to_string() + "/packages/" + package_name;
+
         // Check if package was symlinked
         for package_version in &installed_versions {
             if package_version.symlinked {
-                //TODO
-                cli::display_warning("Package was symlinked, symlink deletion is not yet implemented!");
+                self.remove_symlinks(Path::new(&self.config.prefix_directory), Path::new(&directory))?;
             }
         }
```
BraveChicken1, line 245: Shouldn't we break the loop after we found the first symlinked package? Alternatively we could check for the specific version sub directories, then we can keep this loop.

src/installer/installer.rs
```diff
@@ -342,6 +341,28 @@ impl<'a> Installer<'a> {
...
+            if file.file_type()?.is_dir() {
+                self.remove_symlinks(&file.path(), destination_dir)?;
+
+                // Remove the directory if it is empty after removing symlinks
+                if fs::read_dir(file.path())?.next().is_none() {
+                    fs::remove_dir(file.path())?;
+                }
+            }
+
+            if file.file_type()?.is_symlink() && fs::read_link(file.path())?.starts_with(destination_dir) {
```
BraveChicken1: I would split file_type into a variable here

src/platforms/symlink.rs
```diff
@@ -21,3 +21,9 @@ pub fn create_symlink(source: &Path, destination: &Path) -> Result<(), std::io::
 pub fn create_symlink(source: &Path, destination: &Path) -> Result<(), std::io::Error> {
     panic!("Cannot create link for target, target is not supported.");
 }
+
+pub fn remove_symlink(symlink: &Path) -> Result<(), std::io::Error> {
+    std::fs::remove_file(symlink)?;
```
BraveChicken1, line 42: Maybe we should check if the path is really a symlink?

BraveChicken1 (changes requested):

src/installer/installer.rs
Max13245: I guess we should break then, since only one version should be linked at a time anyway. We might want to track which version is currently symlinked in the future and then we get rid of the loop entirely.

BraveChicken1 (approved):

PR #12, opened 2026-01-03
Add support for dependency versions

Max13245:
This PR adds support for dependency version specification. In order to do this the versions are now represented as a Version struct instead of a version string.


src/cli/commands/list.rs
```diff
@@ -13,7 +13,7 @@ use crate::{
 #[derive(Args, Debug)]
 pub struct ListArgs {
     /// Directory to list all packages of (OPTIONAL)
-    directory: Option<PathBuf>, // TODO: Unused atm
+    directory: Option<PathBuf>, // TODO: Not implemented yet, for when language package support exists
```
BraveChicken1, line 16: I would remove this option for now, but that's probably something for a different commit.

src/cli/commands/search.rs
```diff
@@ -69,12 +70,13 @@ impl HandleCommand for SearchArgs {
             },
         };
 
-        let dependencies: Vec<_> = package_version.dependencies.iter().chain(target.dependencies.iter()).map(|x| x.as_str()).collect();
+        let dependencies: Vec<_> = package_version.dependencies.iter().chain(target.dependencies.iter()).collect();
+        let dependencies: Vec<String> = dependencies.iter().map(|d| d.to_string()).collect();
```
BraveChicken1, line 75: We should probably move these dependency additions into a function, since we need this multiple times.

src/dependency.rs
```diff
@@ -0,0 +1,175 @@
...
+
+#[derive(Debug, Clone)]
+enum VersionBounds {
+    Range(Version, Version),
+    Lower(Version),
+    LowerEqual(Version),
+    Higher(Version),
+    HigherEqual(Version),
+    Equal(Version),
+}
```
BraveChicken1: We could split `VersionBounds` into it's own module, with a `FromStr` implementation for the `parse_version_range` function.

We could then also put the `dependency`, `version` and the new `versionbound` module into a `types` module inside the `installer` module.

PR #13, opened 2026-01-03
Refactor builder into separate module

BraveChicken1:
This PR splits the builder from the installer into its own module. An initial framework for the prebuild packages is also added.

### DIFF
--- a/src/installer/builder.rs
+++ b/src/installer/builder.rs
@@ -7,8 +7,7 @@ use crate::{
     config::Config,
     installed_packages::InstalledPackageStorage,
     installer::{
-        scripts::{self, ScriptError, SCRIPT_EXTENSION},
-        types::Version,
+        scripts::{self, ScriptError},
         unpack::unpack,
     },
     platforms::TARGET_ARCHITECTURE,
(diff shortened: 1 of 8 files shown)


src/installer/builder.rs
```diff
@@ -0,0 +1,141 @@
...
+    TargetError,
+
+    #[error("Cannot request files for building: {0}")]
+    RequestError(#[from] reqwest::Error),
+
+    #[error("Cannot unpack response: {0}")]
+    UnpackError(#[from] std::io::Error),
+
+    #[error("Dependency '{package_name}' of type '{dependency_type}' is not installed.")]
+    DependencyError {
```
Max13245: There are now 2 DependencyError variants, here in the BuilderError and also in the InstallerError, but they do different things. We should probably rename this one to something like "DependencyDeletionError".

src/installer/builder.rs
```diff
@@ -0,0 +1,141 @@
...
+
+        // Finish download spinner
+        spinner.finish("Downloading ".to_string() + &package.name + " successful");
+
+        // Unpack the package to the temp directory
+        let unpack_directory = self.config.temp_directory.join(&package.name).join(package_version.version.to_string());
+        unpack(bytes, &unpack_directory)?;
+
+        // Construct args for the build script
+        let args = package_version
```
Max13245: We now have this in two places, installer and builder. Maybe we could move this to PackageVersion instead and call it with some method "get_combined_args".

src/installer/builder.rs
BraveChicken1: I renamed `DependencyError` in `BuilderError` to `MissingDependencyError`.

PR #14, opened 2026-01-04
Refactor paths

Max13245:
Uses PathBufs instead of strings for paths.

### DIFF
--- a/src/installer/installer.rs
+++ b/src/installer/installer.rs
@@ -212,7 +212,7 @@ impl<'a> Installer<'a> {
             directory = self.config.prefix_directory.join("packages").join(package_name);
         } else {
             // The remove directory of a specific package version
-            directory = self.config.prefix_directory.join("packages").join(package_name);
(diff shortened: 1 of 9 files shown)


src/installer/installer.rs
```diff
@@ -206,13 +207,12 @@ impl<'a> Installer<'a> {
...
-        // TODO: Use join instead
-        let directory: String;
+        let directory: PathBuf;
         if installed_versions.len() == 1 {
-            directory = self.config.prefix_directory.to_string() + "/packages/" + package_name;
+            directory = self.config.prefix_directory.join("packages").join(package_name);
         } else {
             // The remove directory of a specific package version
-            directory = self.config.prefix_directory.to_string() + "/packages/" + package_name + "/" + &version.to_string();
+            directory = self.config.prefix_directory.join("packages").join(package_name);
```
BraveChicken1: The version seems to be missing in this path, please add `.join(version)`

BraveChicken1 (changes requested):
Looks good, just one small bug.

BraveChicken1 (approved):

PR #16, opened 2026-01-24
Refactor display

Max13245:
Changes to be added:
- Correct status codes for exits
- Adjust clap errors


src/installer/types/dependency.rs
```diff
@@ -7,10 +7,10 @@ use crate::installer::types::{Version, VersionBounds, VersionError};
 
 #[derive(Error, Debug)]
 pub enum DependencyParserError {
-    #[error("Cannot parse version number. {0}")]
+    #[error("Caused by: Cannot parse version number\nCaused by: {0}")]
```
BraveChicken1: This first `Caused by` seems to be wrong here.

src/main.rs
```diff
@@ -12,10 +13,5 @@ mod verifier;
 fn main() {
     let config = Config::from(&Config::get_default_path()).expect("Cannot load config");
     let manager = RepositoryManager::new(&config);
-    let cli = commands::Cli::get_instance();
-
-    match cli.handle_command(&manager, &config) {
-        Ok(_) => {},
-        Err(e) => println!("An error occured: {}\n{:?}", e, e),
-    };
+    commands::Cli::get_instance().handle_command(&manager, &config);
```
BraveChicken1: We should probably just import Cli and then we can remove the `commands::`.

src/repositories/types.rs
```diff
@@ -25,6 +28,12 @@ pub struct Package {
     pub latest_versions: HashMap<String, Version>,
 }
 
+impl Package {
+    pub fn get_latest_version(&self, target_name: &str) -> Result<&Version, RepositoryError> {
```
BraveChicken1: These results could use repository::error::Result, then we don't need to specify the RepositoryError here.

BraveChicken1 (commented):
Looks good overall, just a few comments.
For the unwrap_or_exit: don't we need to be able to specify an error message to be printed before exiting?

PR #20, opened 2026-02-02
Refactor installed toml

Max13245:
- Split (and rename) the installed packages (register) into different parts and use hashmaps for quick indexing.
- Removes external fields, because the checks don't exist yet.

Some renaming changes:
- Package register --> Overview of all package names
- InstalledPackage --> Information about a package and an overview of all package versions
- InstalledPackageVersion --> All version specific package information
- PackageMetadata --> Previously Package, it holds package metadata defined in the repositories


src/installer/installer.rs
```diff
@@ -210,73 +215,56 @@ impl<'a> Installer<'a> {
...
-        let directory: PathBuf;
-        if installed_versions.len() == 1 {
-            directory = self.config.prefix_directory.join("packages").join(package_name);
-        } else {
-            // The remove directory of a specific package version
-            directory = self.config.prefix_directory.join("packages").join(package_name).join(version.to_string());
-        }
+        let installed_package = match self.register.packages.get(&package_id.name) {
+            Some(package) => package,
+            None => unreachable!("Should be unreachable, checked for package version before"),
```
BraveChicken1: I would change this into a warning with a return, just to be sure.

src/installer/installer.rs
```diff
@@ -210,73 +215,56 @@ impl<'a> Installer<'a> {
...
-            other_versions.retain(|x| x.version != *version);
-            other_versions.sort_by_key(|x| &x.version);
+        if installed_package.active == package_id.version {
+            let mut other_versions = self.register.get_package_versions(&package_id.name);
+            other_versions.retain(|x| x.package_id.version != package_id.version);
+            other_versions.sort_by_key(|x| &x.package_id.version);
 
             if let Some(newest) = other_versions.last() {
-                self.set_active(package_name, &newest.version.clone(), installed_package.symlinked)?;
+                // TODO: Look at clone
```
BraveChicken1: Can you not just borrow here?

src/installer/installer.rs
```diff
@@ -472,22 +459,14 @@ impl<'a> Installer<'a> {
...
-            }
-
-            // Update other packages in storage
-            installed.active = false;
-            installed.symlinked = false;
+        // Updates the active version and sets its symlinked state
+        if let Some(package) = self.register.packages.get_mut(&package_id.name) {
+            package.active = package_id.version.clone();
+            package.symlinked = should_symlink;
         }
```
BraveChicken1, line 493: We should probably error when we cannot find something here, since it would result in an invalid register.

src/storage/installed_package.rs
```diff
@@ -0,0 +1,64 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::{installer::types::Version, repositories::types::PackageMetadata, storage::installed_package_version::InstalledPackageVersion};
+
+/// The installed package holds package specific info and a mapping from versions to package versions.
+#[derive(Deserialize, Serialize, Debug)]
+pub struct InstalledPackage {
+    pub versions: HashMap<Version, InstalledPackageVersion>,
+    pub symlinked: bool,
+    pub active: Version,
```
BraveChicken1: I would rename this field to `active_version` to make clearer its a version.

src/storage/mod.rs
```diff
@@ -0,0 +1,5 @@
+pub mod error;
+pub mod installed_package;
+pub mod installed_package_version;
+//pub mod installed_packages;
```
BraveChicken1: Please remove this commented code

src/storage/package_register.rs
```diff
@@ -0,0 +1,248 @@
...
+    storage::{error::InstalledPackagesError, installed_package::InstalledPackage, installed_package_version::InstalledPackageVersion},
+    utils::constants::INSTALLED_FILENAME,
+};
+
+const HEADER_TEXT: &str = "# DO NOT EDIT THIS FILE";
+
+/// The package register holds references to packages with package names.
+#[derive(Deserialize, Serialize, Debug)]
+pub struct PackageRegister {
+    pub packages: HashMap<String, InstalledPackage>,
```
BraveChicken1: I would remove the pub here, and add functions to access it instead. This prevents strange issues when mutating the HashMap directly.

src/storage/package_register.rs
```diff
@@ -0,0 +1,248 @@
...
+
+        // Write data to file
+        fs::write(path, data)?;
+
+        Ok(())
+    }
+
+    /// Gets the default path of the Packit installed packages file.
+    pub fn get_default_path() -> PathBuf {
+        Path::new(DEFAULT_CONFIG_DIR).join(INSTALLED_FILENAME)
```
BraveChicken1: Please rename `INSTALLED_FILENAME` to `REGISTER_FILENAME`.

src/storage/package_register.rs
```diff
@@ -0,0 +1,248 @@
...
+            None => package_version.dependencies.iter().collect(),
+        };
+
+        // Get the satisfying package for the dependencies
+        let mut dependency_ids = HashSet::new();
+        for dependency in dependencies {
+            if let Some(package) = self.get_satisfying_package(dependency) {
+                dependency_ids.insert(package.package_id.clone());
+            }
+        }
```
BraveChicken1: Doesn't the Installer normally decide which dependency versions are installed? Shouldn't we get the dependencies as a parameter then?

src/storage/package_register.rs
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    /// Removes a package version from the register storage.
+    /// Please note that this does not save the storage and does not read the currently installed packages from the toml.
+    pub fn remove_package_version(&mut self, package_id: &PackageId) {
+        // Remove and get the package, if it doesn't exist return early.
+        let mut package = match self.packages.remove(&package_id.name) {
+            Some(package) => package,
+            None => return,
+        };
```
BraveChicken1: We shouldn't remove a package when removing a specific version right?
This should only happen when the removed version was the only installed version of the package.

src/storage/package_register.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    /// Checks if a certain package is a dependency by checking its dependents.
+    /// If only a package name is given we check all installed versions of that
+    /// package, if any of those are a dependency we return true.
+    pub fn is_dependency(&self, package_name: &str, version: Option<&Version>) -> bool {
+        // Only check a specific package version if a version is provided
+        if let Some(version) = version {
+            return match self.get_package(&PackageId::new(package_name, version)) {
+                Some(package) => !package.dependents.is_empty(),
+                None => false, // If the package doesn't exist it's not a dependency
```
BraveChicken1: This should probably print a warning, since it would be strange behaviour.

BraveChicken1 (changes requested):
Looks good overall, just a few comments.

src/installer/installer.rs
Max13245: No I can't mutably borrow there because self is borrowed a couple lines earlier by:
```let mut other_versions = self.register.get_package_versions(&package_id.name);```

src/storage/package_register.rs
Max13245: Yes you're completely right, fixing it now :)

src/storage/package_register.rs
Max13245: This could be possible, but then the install_dependency method in the installer should return the dependency ids, because we need to know the specific dependency versions.

src/installer/installer.rs
BraveChicken1: Then we have to keep the clone, you can remove the TODO then.

src/storage/package_register.rs
BraveChicken1: That would probably be neater to do, because now we could get unwanted behaviour

Max13245: okido, wil fix that then :)

BraveChicken1 (approved):

PR #21, opened 2026-02-05
Add extra commands

BraveChicken1:
This pull requests implements extra commands and refactors the install command to allow installing multiple packages with a single command.

New commands:
|                                           |                                               |
| ------------------------ | --------------------------  |
| install --build                    | Build package from source |
| install --skip-symlinking | Skip symlinking of the newly installed package |
| install --skip-active         | Skip setting newly installed package to active |
| switch                               | Switch the current active version of a package |
| link and unlink                  | Switch the symlinked state of a package |
| fix                                      | Fix the packit installation, such as correctly setting permissions and repairing broken  register entries. Implementation will be done later. |

### DIFF
--- a/src/cli/commands/link.rs
+++ b/src/cli/commands/link.rs
@@ -1,0 +1,100 @@
+use clap::Args;
+
+use crate::{
+    cli::{commands::HandleCommand, display::logging::warning},
+    config::{Config, Repository},
+    error_handling::HandleError,
+    installer::installer::Installer,
+    platforms::TARGET_ARCHITECTURE,
+    repositories::{manager::RepositoryManager, provider},
+    storage::package_register::PackageRegister,
+};
+
+#[derive(Args, Debug)]
+pub struct LinkArgs {
+    /// The name of the package to link
+    pub package_name: String,
+
+    /// True to force linking, even when we should not link
+    #[arg(short, long, default_value = "false")]
+    pub force: bool,
+}
+
+impl HandleCommand for LinkArgs {
+    fn handle(&self, config: &Config, manager: &RepositoryManager) {
+        let register_path = PackageRegister::get_default_path();
+        let mut register = PackageRegister::from(&register_path).unwrap_or_exit(1);
+
+        // Get installed package
+        let package = register
+            .get_package(&self.package_name)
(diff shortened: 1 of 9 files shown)


src/cli/commands/link.rs
```diff
@@ -0,0 +1,100 @@
...
+                println!("The package metadata defines we should not symlink this package, cancelling linking.");
+                return;
+            }
+        }
+
+        let install_path = package_version.install_path.clone();
+
+        // Create symlinks
+        Installer::new(config, &mut register, manager)
+            .create_symlinks(&install_path)
```
Max13245, line 88: Maybe it's nice to split the (un)link functionality from the installer, since it is now also used outside of it.

src/installer/installer.rs
```diff
@@ -174,7 +188,8 @@ impl<'a> Installer<'a> {
             // Determine the latest supported version for the dependency
             let version = self.get_latest_dependency_version(dependency)?;
 
-            self.install(dependency.get_name(), Some(&version))?;
+            // TODO: should we pass the options to childs? We could add these options as struct values
```
Max13245, line 191: I would say that the options should also apply to the children.

src/installer/installer.rs
```diff
@@ -497,4 +512,50 @@ impl<'a> Installer<'a> {
...
+            let entry = entry?;
+
+            if entry.file_type()?.is_dir() && entry.file_name() != "active" {
+                self.remove_symlinks(&entry.path(), &install_path)?;
+            }
+        }
+
+        // Update symlinked state in package register
+        match self.register.get_package_mut(&package_name) {
+            Some(package) => package.symlinked = true,
```
Max13245: Should this bool not be set to false instead?

src/cli/commands/link.rs
BraveChicken1: I will address this in a separate PR.

src/installer/installer.rs
BraveChicken1: I will address this in a separate PR, together with the other comment.

Max13245 (approved):

PR #23, opened 2026-02-07
Refactor installer

BraveChicken1:
This PR refactors the installer by splitting options into `InstallerOptions` and splitting symlink related functions into a new `Symlinker`.

### DIFF
--- a/src/installer/installer.rs
+++ b/src/installer/installer.rs
@@ -176,7 +176,7 @@ impl<'a> Installer<'a> {
         // Get build version of package
         match url {
             Some(url) => self.download_prebuild(&url, &install_directory)?,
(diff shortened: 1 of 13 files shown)


src/installer/installer.rs
```diff
@@ -30,6 +32,7 @@ pub struct Installer<'a> {
     config: &'a Config,
     register: &'a mut PackageRegister,
     repository_manager: &'a RepositoryManager<'a>,
+    options: InstallerOptions,
```
Max13245, line 35: If eventually we want to give options to the uninstall command as well, we might want to give this as a parameter to the install method instead.

Max13245:
Overall it looks good!

PR #25, opened 2026-02-09
Implement grouped metadata

BraveChicken1:
This PR implements grouped metadata.

### DIFF
--- a/src/platforms/target_architecture.rs
+++ b/src/platforms/target_architecture.rs
@@ -30,3 +30,23 @@ pub const TARGET_ARCHITECTURE: &str = "x86_64-pc-windows-msvc";
 #[cfg(target_arch = "aarch64")]
 #[cfg(target_env = "msvc")]
 pub const TARGET_ARCHITECTURE: &str = "aarch64-pc-windows-msvc";
+
+pub fn get_os_name(target_name: &str) -> &str {
+    match target_name {
+        "x86_64-apple-darwin" | "aarch64-apple-darwin" => "mac",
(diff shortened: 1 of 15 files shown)


src/platforms/target_architecture.rs
```diff
@@ -30,3 +30,23 @@ pub const TARGET_ARCHITECTURE: &str = "x86_64-pc-windows-msvc";
 #[cfg(target_arch = "aarch64")]
 #[cfg(target_env = "msvc")]
 pub const TARGET_ARCHITECTURE: &str = "aarch64-pc-windows-msvc";
```
Max13245, line 32: Maybe create an enum for the architectures.

src/repositories/types/package_version.rs
```diff
@@ -138,6 +160,24 @@ impl PackageVersionMeta {
...
+        if let Sources::Named(sources) = &self.sources {
+            for (_, target) in &self.targets {
+                match &target.source {
+                    Some(source) if !sources.contains_key(source) => return true,
+                    None => return true,
+                    _ => continue,
+                }
+            }
+        }
+
```
Max13245, line 180: If we update our rust version we could maybe do something like this instead:

```
if let Sources::Named(sources) = &self.sources {
            for (_, target) in &self.targets {
                if let Some(source) = &target.source && sources.contains_key(source)  {
                    continue;
                }

                return true;
            }
        }
```

Max13245: sry for the wonky spacing, github won't let me edit it...

Max13245 (approved):

PR #28, opened 2026-02-13
Implement uninstall scripts

BraveChicken1:
This PR implements uninstall scripts, which can be used to cleanup at uninstall time.

### DIFF
--- a/src/installer/installer.rs
+++ b/src/installer/installer.rs
@@ -1,10 +1,10 @@
 use crate::{
     cli::display::{
         ask_user,
-        logging::{debug, warning},
+        logging::{debug, error, warning},
         QuestionResponse,
     },
-    config::Config,
+    config::{Config, Repository},
     installer::{
         builder::Builder,
         error::{InstallerError, Result},
@@ -16,6 +16,7 @@ use crate::{
     platforms::{symlink, TARGET_ARCHITECTURE},
     repositories::{
         manager::RepositoryManager,
+        provider,
         types::{PackageMeta, PackageTarget, PackageVersionMeta},
     },
     storage::package_register::PackageRegister,
@@ -435,6 +436,18 @@ impl<'a> Installer<'a> {
             },
         };
 
+        // Load source repository
+        let repository = match installed_package.get_package_version(&package_id.version) {
+            Some(package_version) => Repository::new(&package_version.source_repository_url, &package_version.source_repository_provider),
+            None => {
+                warning!("Package version not found eventhough package version was found, should be unreachable.");
+                return Ok(());
+            },
+        };
+
+        // Run uninstall script
+        self.run_uninstall_script(&repository, package_id, &directory)?;
+
         // Check if the package was symlinked
         if installed_package.active_version == package_id.version && installed_package.symlinked {
             Symlinker::new(self.config).remove_symlinks(Path::new(&self.config.prefix_directory), Path::new(&directory))?;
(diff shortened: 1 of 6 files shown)


src/installer/installer.rs
```diff
@@ -435,6 +436,18 @@ impl<'a> Installer<'a> {
             },
         };
 
+        // Load source repository
+        let repository = match installed_package.get_package_version(&package_id.version) {
+            Some(package_version) => Repository::new(&package_version.source_repository_url, &package_version.source_repository_provider),
+            None => {
+                warning!("Package version not found eventhough package version was found, should be unreachable.");
```
Max13245: Please change the warning message to:
"Package version cannot be found eventhough it was found before, should be unreachable."

src/installer/installer.rs
Max13245, line 435: Please also change the warning message here to:
"Package version cannot be found eventhough it was found before, should be unreachable."

src/installer/installer.rs
```diff
@@ -504,6 +526,41 @@ impl<'a> Installer<'a> {
         Ok(())
     }
 
+    fn run_uninstall_script(&self, repository: &Repository, package_id: &PackageId, install_directory: &PathBuf) -> Result<()> {
+        // Create repository provider for source repository
+        let provider = match provider::create_repository_provider(&repository) {
+            Some(provider) => provider,
+            None => {
+                warning!("Unable to create repository provider to retrieve uninstall script");
```
Max13245: I think this should be an error, because, although unlikely, it is not unreachable. If a user changed the source repository field manually for some reason they would get a warning here, but also a success statuscode. Making it seems like the uninstall was successful, eventhough it wasn't. Please make this an error.

BraveChicken1: I agree this should be an error, but that would mean we wouldn't be able to uninstall packages when the source repository does not exist anymore.
Imagine we installed a package from some repository, and then later the repository is not reachable anymore, that would lead to the situation where the uninstall is not possible anymore.

An alternative to this situation would be to locally cache the repository metadata and scripts when installing a package.

Max13245: Mhmm good point. If we cache it, we could maybe even only do it when a user uninstalls a repository provider. When that happens we check which packages are installed from that provider, then we store the metadata for those only. Then we should also have a check when installing a new provider if there are already packages installed from that provider, so we can remove the stored metadata. I don't know if this is too much of a hassle though. Are there other situations where we would need this cache? If yes, then we could just do it always.

Max13245: Although, I just realise, this doesn't account for the situation where the provider isn't reachable anymore, because of some external factor.

BraveChicken1: We need to access package metadata after installation in several places. For caching I would suggest doing the following:
After installation of a package we cache the targets.toml file and the uninstall script. If needed we could also cache other scripts. We store these files somewhere in a local cache and keep them until the package is uninstalled. If we then need to access the metadata somewhere after the package was installed, we can simply pull it from the cache. Optionally we could try fetching from the repository first and only use the cache if we cannot find the repository anymore.

Max13245: That seems like a good idea. I would suggest that we use the cache first (if possible), because if there was a revision of the metadata it could change how we uninstall the package. If we then used the remote version it might not be compatible with the current local version.

BraveChicken1: That's true, however we have to then discuss what we should do if we need the revision to fix the uninstall script

BraveChicken1: I changed the message to log as an error for now, I will implement caching in a different PR later.

Max13245 (approved):

PR #29, opened 2026-02-13
Refactor command package versions

BraveChicken1:
This PR refactors the package version argument in the uninstall and search command by implementing a new `OptionalPackageId`. The OptionalPackageId is also implement in several functions in the installer.

### DIFF
--- a/src/installer/installer.rs
+++ b/src/installer/installer.rs
@@ -11,7 +11,7 @@ use crate::{
         options::InstallerOptions,
         scripts::{self, ScriptData},
         symlinker::Symlinker,
-        types::{Dependency, PackageId, Version},
+        types::{Dependency, OptionalPackageId, PackageId, Version},
     },
     platforms::{symlink, TARGET_ARCHITECTURE},
     repositories::{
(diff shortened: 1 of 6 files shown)


src/installer/types/package_id.rs
```diff
@@ -82,6 +82,75 @@ impl FromStr for PackageId {
...
+            });
+        }
+
+        Ok(Self {
+            name: string.into(),
+            version: None,
+        })
+    }
+}
+
```
Max13245, line 124: Do we really need to deserialize here? Because we don't want OptionalPackageId's stored. I think, at the point of storage, we should always be sure which version we have and there should be an error (caused by parsing from the PackageId) if the version isn't specified.

src/installer/installer.rs
```diff
@@ -371,10 +371,10 @@ impl<'a> Installer<'a> {
                 continue;
             }
```
Max13245, line 374: I think here we might actually want to use the uninstall_single method instead. Then we don't need to create an OptionalPackageId. Then you might want to add the permission check in the remove_build_dependencies method (although there is a check in the install method which calls it).

Max13245 (commented):
OptionalPackageId are good for user input where we just can't always be sure of the specific package version, but I think it should really be limited to user input. So we might want to delete the new methods, then an OptionalPackageId can only come from parsing it from the user input.

src/installer/types/package_id.rs
BraveChicken1: I'll remove the deserialize since it's indeed not needed here.

src/installer/installer.rs
BraveChicken1: I wouldn't call the uninstall_single method directly, that would mean we would get duplicate code in the uninstall_all and uninstall_single command. An alternative could be to remove the OptionalPackageId here.

BraveChicken1:
We could remove the new methods, as an alternative we could then implement the From PackageId trait, to support converting a PackageId to OptionalPackageId

Max13245:
The reason I suggest we remove the "new" methods is that we should only use OptionalPackageId structs when there is user input and we can't be sure about the version. In all other cases we shouldn't be using the OptionalPackageId, because there is probably a way to know what the version should be. If in the future we see there is a case where we still need these "new" method or like you suggested the From trait for the PackageId then we could always add it then, when we are sure it's necessary.

BraveChicken1:
In my implementation I added the OptionalPackageId to the uninstall and install functions too, because we determine the version inside these functions. That means we do need the new methods to call the uninstall method from within the Installer.

Max13245:
Mhmm, okay that's true. Let's go with the From trait implementation then and I do think we should be careful to only use it when necessary.

Max13245 (approved):

PR #32, opened 2026-02-18
Add installing of prebuilds

BraveChicken1:
This PR adds the installing of prebuild packages. It implements the new PrebuildProvider and the actual downloading in the installer.


src/repositories/error.rs
```diff
@@ -23,11 +26,21 @@ pub enum RepositoryError {
         version: Option<String>,
     },
 
+    #[error("Cannot find prebuild of package '{package_name}@{version}' revision {revision}")]
```
Max13245: The PackageId object implements the Display trait, which also formats it like: <package-name>@<version>. So you could pass the package id directly instead.

Max13245 (approved):

PR #33, opened 2026-02-19
Implement fix and check command

Max13245:
This PR adds a couple of things:
- It fixes inconsistent metadata in tar headers packages.
- It adds a check command to scan the Packit environment for issues.
- It adds a fix command which will fix issues found in the Packit environment.

### DIFF
--- a/src/installer/symlinker.rs
+++ b/src/installer/symlinker.rs
@@ -9,7 +9,7 @@ use crate::{
     },
     platforms::symlink,
     storage::package_register::PackageRegister,
-    utils::symlink::{create_folder_symlinks, remove_symlinks},
+    utils::io::{create_folder_symlinks, remove_symlinks},
 };
 
 pub struct Symlinker<'a> {
--- a/src/installer/symlinker.rs
+++ b/src/installer/symlinker.rs
@@ -9,6 +9,7 @@ use crate::{
     },
     platforms::symlink,
     storage::package_register::PackageRegister,
+    utils::symlink::{create_folder_symlinks, remove_symlinks},
 };
 
 pub struct Symlinker<'a> {
@@ -28,76 +29,14 @@ impl<'a> Symlinker<'a> {
             let package_dir_path = package_directory.join(dir_name);
             let prefix_dir_path = prefix_dir.join(dir_name);
 
-            self.create_folder_symlinks(&package_dir_path, &prefix_dir_path, true)?;
+            create_folder_symlinks(&package_dir_path, &prefix_dir_path, true)?;
         }
 
         Ok(())
     }
 
-    fn create_folder_symlinks(&self, source_dir: &Path, destination_dir: &Path, keep_subdirectories: bool) -> Result<()> {
-        // Create destination if it does not exist
-        if !destination_dir.exists() {
-            fs::create_dir_all(&destination_dir)?;
-        }
-
-        // Skip symlinking if source does not exist
-        if !source_dir.exists() {
-            return Ok(());
-        }
-
-        // Symlink files
-        for file in fs::read_dir(source_dir)? {
-            let file = file?;
-
-            let destination = destination_dir.join(file.file_name());
-
-            // Handle directories
-            if file.file_type()?.is_dir() {
-                // If we want to keep subdirectories, create the symlinks for the subdirectory
-                // TODO: Handle subdirectories properly
-                if keep_subdirectories {
-                    self.create_folder_symlinks(&file.path(), &destination, true)?;
-                } else {
-                    dbg!("Skipping subdirectory", file);
-                }
-
-                continue;
-            }
-
-            // Check if file already exists
-            if fs::exists(&destination)? {
-                warning!("Symlink {:?} already exists in {:?}", file.file_name(), destination_dir);
-                continue;
-            }
-
-            // Symlink file in destination directory
-            symlink::create_symlink(&file.path(), &destination)?;
-        }
-
-        Ok(())
-    }
-
-    /// Searches for symlinks with a certain destination (destinations inside of the destination are also a match).
-    pub fn remove_symlinks(&self, search_dir: &Path, destination_dir: &Path) -> Result<()> {
-        for file in fs::read_dir(search_dir)? {
-            let file = file?;
-            let file_type = file.file_type()?;
-
-            if file_type.is_dir() {
-                self.remove_symlinks(&file.path(), destination_dir)?;
-
-                // Remove the directory if it is empty after removing symlinks
-                if fs::read_dir(file.path())?.next().is_none() {
-                    fs::remove_dir(file.path())?;
-                }
-            }
-
-            if file_type.is_symlink() && fs::read_link(file.path())?.starts_with(destination_dir) {
-                symlink::remove_symlink(&file.path())?
-            }
-        }
-
-        Ok(())
+    pub(super) fn remove_symlinks(&self, search_dir: &Path, destination_dir: &Path) -> Result<()> {
+        Ok(remove_symlinks(search_dir, destination_dir)?)
     }
 
     /// Sets a package to active and create the appropiate symlinks for it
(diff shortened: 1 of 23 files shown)


src/installer/symlinker.rs
```diff
@@ -78,7 +78,7 @@ impl<'a> Symlinker<'a> {
     }
 
     /// Searches for symlinks with a certain destination (destinations inside of the destination are also a match).
-    pub(super) fn remove_symlinks(&self, search_dir: &Path, destination_dir: &Path) -> Result<()> {
+    pub fn remove_symlinks(&self, search_dir: &Path, destination_dir: &Path) -> Result<()> {
```
BraveChicken1: This function was private within the installer module because it shouldn't be called from outside, unlink_package is meant for removing the symlinks of a package in a clean way.

src/verifier/mod.rs
```diff
@@ -0,0 +1,10 @@
+mod error;
```
BraveChicken1: Shouldn't the error module be public?

src/verifier/repairer.rs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+impl<'a> Repairer<'a> {
+    /// Creates a new repairer.
+    pub fn new(config: &'a Config, manager: &'a RepositoryManager) -> Self {
+        Self { config, manager }
+    }
+
+    /// Gets the issues one by one with the verifier and fixes them.
+    pub fn fix(&mut self, verifier: &mut Verifier, register: &mut PackageRegister) -> Result<(), VerifierError> {
```
BraveChicken1: You could change Result<(), VerifierError> into Result<()>. In the error module you could then add a Result<T, VerifierError>.

src/verifier/repairer.rs
```diff
@@ -0,0 +1,123 @@
...
+        while let Some(issue) = verifier.next_issue(register)? {
+            print!("{issue}\n");
+
+            let question = "Would you like to automatically fix the above issue with `pit fix`?";
+            if ask_user(question, QuestionResponse::Yes)?.is_no() {
+                continue;
+            }
+
```
BraveChicken1: Shouldn't this be extracted into the fix command? The function in the repairer would then simply be fix(issue)

src/verifier/repairer.rs
```diff
@@ -0,0 +1,123 @@
...
+        Ok(())
+    }
+
+    /// Fixes broken dependency trees by installing the missing packages.
+    fn fix_broken_tree(&mut self, missing: Vec<(PackageId, PackageId)>, register: &mut PackageRegister) -> Result<(), VerifierError> {
+        let installer_options = InstallerOptions::default().skip_symlinking(true);
+        let mut installer = Installer::new(&self.config, register, &self.manager, installer_options);
+
+        for (_, missing_package) in missing {
+            // TODO: Do we want to check for already existing packages? (there could be duplicates in the missing dependencies)
```
BraveChicken1: We should probably check for installed package that already satisfy the dependency, but that would be difficult since we can only get that from the package metadata. Maybe something to implement when we have metadata caching.

src/verifier/repairer.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    /// Fixes inconsistent storage by temporarily removing the missing package from the register and then re-installing the packages.
+    fn fix_inconsistent_storage(&mut self, missing: Vec<PackageId>, register: &mut PackageRegister) -> Result<(), VerifierError> {
+        for missing_package in missing {
+            // Gather the package settings before removing the package from the register
+            let (symlinked, active) = match register.get_package(&missing_package.name) {
+                Some(package) => (package.symlinked, package.active_version == missing_package.version),
+                None => {
+                    warning!("Inconsistent package cannot be found in Installed.toml anymore, eventhough it could be found before.");
```
```suggestion
                    warning!("Inconsistent package cannot be found in Installed.toml anymore, eventhough it was found before.");
```

src/verifier/repairer.rs
```diff
@@ -0,0 +1,123 @@
...
+
+            // Figure out if symlinked
+            let symlinked = !fs::symlink_metadata(bin_directory.join(&package_id.name)).is_err();
+
+            // Temporarily remove the package from the storage
+            // We have to uninstall and install, because we can't know the repository source otherwise
+            // Remove the symlinks first
+            // TODO: Use uninstall scripts here in the future
+            if symlinked {
+                Symlinker::new(self.config).remove_symlinks(Path::new(&self.config.prefix_directory), Path::new(&package_directory))?;
```
BraveChicken1: This should probably use the unlink_package instead of remove_symlinks

src/verifier/verifier.rs
```diff
@@ -0,0 +1,266 @@
...
+
+    /// Checks for alterations in a single package using a checksum which is compared to the checksum from the pre-build.
+    /// Returns true if the package was altered, false if not.
+    fn check_package_alterations(&self, package_id: &PackageId) -> Result<bool, VerifierError> {
+        let install_directory = self.config.prefix_directory.join("packages").join(&package_id.name).join(package_id.version.to_string());
+        let mut compressed = packager::compress(&install_directory)?;
+        let checksum = Checksum::calculate_checksum(&mut compressed)?;
+
+        // TODO: Actually search for the correct checksum
+        let correct_checksum = "904723e939e42c4cfa16b2cc8bf9f7d637a6f56b568975b0b97cb897ce5fff15";
```
BraveChicken1: Wouldn't it be more efficient to compare checksums as a byte array, instead of as string?

src/packager.rs
```diff
@@ -18,45 +20,160 @@ use crate::{
...
         return Err(io::Error::new(io::ErrorKind::NotADirectory, "Destination is not a directory."))?;
     }
 
-    // Create pre-package filename
-    let filename = format!("{package_id}-{revisions}-{TARGET_ARCHITECTURE}.tar.gz");
+    // Compress the package
+    let mut compressed = compress(&install_directory)?;
+
+    // Calculate checksum for the compressed
+    let checksum = Checksum::calculate_checksum(&mut compressed)?;
```
BraveChicken1: Maybe we should return a Checksum struct from the calculate_checksum method and then implement to_string for Checksum

BraveChicken1 (changes requested):
Looks good overall, just some small comments.

src/verifier/repairer.rs
Max13245: The problem here is that the missing vec can contain duplicates. Which means that the first time it is installed it's okay, but the second time it will say package X is already installed. Which is a bit weird, but when I try to check for it with the register I get borrowing issues. Alternatives would be to either, remove the duplicates or keep a seen HashSet.

src/verifier/repairer.rs
Max13245: Yes, that's also fine.

src/verifier/verifier.rs
Max13245: Yes, comparing strings has a bit more overhead. I'll change it!

src/packager.rs
Max13245: For now I just removed the implementation of the struct, because it would be really small.

src/verifier/repairer.rs
Max13245: I agree that remove_symlinks should be private, however that would mean complete duplicate logic of that method, because there the package is uninstalled from storage (not the register). 

Maybe it is possible to split the uninstall from storage en register in the installer. Then I don't have to do it here, but that's also not great.

src/verifier/repairer.rs
BraveChicken1: I think keeping a seen HashSet is then probably the cleanest way.

src/verifier/repairer.rs
BraveChicken1: The unlink_package does not uninstall the package, it only removes the symlinks in a clean way. Using that here shouldn't be a problem.

src/packager.rs
BraveChicken1: A to_string method is probably neat for writing the checksum to the file.

Max13245: Sure I could do that, but it would just be a wrapper around hex::encode.

src/verifier/repairer.rs
Max13245: Yes, I know it doesn't uninstall the package, but it is part of the uninstall. I need to uninstall the package there (only the package storage not from the register). So I need everything that is in that remove_symlink method, I can call that method like now or I could create an entire duplicate method in which I do the same. Maybe the cleanest way would be to make it a util, in the Symlinker there could still be a method which wraps around it?

src/verifier/repairer.rs
Max13245: I ended up solving the borrowing issue by recreating the installer struct inside of the loop.

BraveChicken1 (approved):

PR #36, opened 2026-02-24
Implement unix permissions

BraveChicken1:
This PR implements proper permissions for unix. Windows permissions are still TODO.

### DIFF
--- a/src/platforms/permissions.rs
+++ b/src/platforms/permissions.rs
@@ -6,11 +6,11 @@ pub enum PermissionError {
     #[error("Error while fetching permissions")]
     IOError(#[from] std::io::Error),
 
-    #[error("User does not exist")]
-    UserDoesNotExist,
-
     #[error("Group does not exist")]
     GroupDoesNotExist,
+
+    #[error("String contains a nul byte")]
+    NulError(#[from] std::ffi::NulError),
 }
 
 type Result<T> = core::result::Result<T, PermissionError>;
@@ -78,7 +78,7 @@ pub mod platform {
             Err(e) => {
                 if matches!(e, PermissionError::GroupDoesNotExist) {
                     warning!("The 'packit' group does not exist. Please run 'pit fix' to fix your Packit installation");
-                    // TODO: pit fix does not exist yet
+                    // TODO: Add group check to verifier
                 }
                 return Err(e);
             },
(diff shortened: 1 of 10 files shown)


src/platforms/permissions.rs
```diff
@@ -0,0 +1,177 @@
...
+            fs::{MetadataExt, PermissionsExt},
+        },
+        path::PathBuf,
+    };
+
+    use crate::cli::display::logging::warning;
+
+    use super::{PermissionError, Result};
+
+    pub const PACKIT_GROUP_NAME: &str = "packit";
```
Max13245, line 53: When windows permissions are also implemented the packit group name should not be specific to a platform. So it should probably be defined in the permissions module.

src/platforms/permissions.rs
```diff
@@ -0,0 +1,177 @@
...
+            set_ownership(path, None, Some(group_id))?;
+
+            old_permissions.set_mode(0o775);
+        } else {
+            old_permissions.set_mode(0o755);
+        }
+
+        fs::set_permissions(&path, old_permissions)?;
+
+        if recurse && path.is_dir() {
```
Max13245: It might be nice to do an early return so that the for loop isn't nested:
```suggestion
        if !recurse || !path.is_dir() {
            return Ok(());
        }

        let dir = fs::read_dir(&path)?;
        for entry in dir {
            let entry = entry?;

            set_file_permissions(&entry.path(), entry.metadata()?.permissions(), group_id, recurse)?;
        }
```

src/platforms/permissions.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    // Use platform specific writable checks
+    platform::is_writable(path, metadata)
+}
+
+pub use platform::set_packit_permissions;
+
+#[cfg(any(target_os = "macos", target_os = "linux"))]
+pub mod platform {
```
Max13245, line 38: I'm not sure about the name platform, it might be a bit broad/vague. Maybe "platform_permissions" or "packit_permissions"?

Max13245 (commented):
Overal it looks really good! These are some very minor remarks.

src/platforms/permissions.rs
BraveChicken1: I am not fully sure if we need the same group functionality on Windows, that's why it's currently on the platform.

src/platforms/permissions.rs
BraveChicken1: I chose platform because it contains the functions for a specific platform. The full path is already `permissions::platform`.

Max13245 (approved):

PR #37, opened 2026-02-28
Add tree module

Max13245:
This PR introduces a generic tree struct to be used for 'package tree' related operations. It implements the display trait for nice display of any tree. It is also implemented in the installer, also removing the flattened dependency tree approach.

This PR also introduces an info command to show package info of installed packages. It accepts a tree parameter for the display of dependency trees.

### DIFF
--- a/src/cli/commands/info.rs
+++ b/src/cli/commands/info.rs
@@ -116,14 +116,14 @@ impl InfoArgs {
         }
         println!();
 
-        print!("Dependent: ");
-        if package_version.dependents.is_empty() {
-            print!("None");
-        }
-
         // Early return if verbose is disabled
         if !self.verbose {
             return;
+        }
+
+        print!("Dependent: ");
+        if package_version.dependents.is_empty() {
+            print!("None");
         }
 
         for dependent in &package_version.dependents {
--- a/src/cli/commands/info.rs
+++ b/src/cli/commands/info.rs
@@ -62,6 +62,7 @@ impl HandleCommand for InfoArgs {
 }
 
 impl InfoArgs {
+    /// Displays package info.
     fn display_package_info(&self, package: &InstalledPackage) {
         println!("{}", self.package.name);
         println!("{}", package.description);
@@ -79,6 +80,7 @@ impl InfoArgs {
         println!("Symlinked: {}", package.symlinked);
     }
 
+    /// Displays the package version info, also checking for the verbose flag for some info.
     fn display_package_version_info(&self, package_id: &PackageId, register: &PackageRegister, package: &InstalledPackage) {
         let package_version = match register.get_package_version(&package_id) {
             Some(package) => package,
--- a/src/cli/commands/info.rs
+++ b/src/cli/commands/info.rs
@@ -8,7 +8,7 @@ use crate::{
     installer::types::{OptionalPackageId, PackageId},
     repositories::manager::RepositoryManager,
     storage::{installed_package::InstalledPackage, package_register::PackageRegister},
-    utils::unwrap_or_exit::UnwrapOrExit,
+    utils::{tree::Node, unwrap_or_exit::UnwrapOrExit},
 };
 
 #[derive(Args, Debug)]
@@ -38,6 +38,18 @@ impl HandleCommand for InfoArgs {
                 exit(1)
             },
         };
+
+        // Display tree if tree flag is given
+        if self.tree {
+            if let Some(package_id) = self.package.versioned() {
+                let tree = Node::new(&package_id, &register).unwrap_or_exit(1);
+                println!("{tree}");
+                return;
+            } else {
+                error!(msg: "Displaying a tree requires package version to be specified.");
+                exit(1)
+            }
+        }
 
         // Show package version specific information
         if let Some(package_id) = self.package.versioned() {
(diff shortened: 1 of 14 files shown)


src/cli/commands/info.rs
```diff
@@ -0,0 +1,148 @@
...
+impl HandleCommand for InfoArgs {
+    fn handle(&self, _: &Config, _: &RepositoryManager) {
+        let register_dir = PackageRegister::get_default_path();
+        let register = PackageRegister::from(&register_dir).unwrap_or_exit(1);
+
+        // Get package information
+        let package = match register.get_package(&self.package.name) {
+            Some(package) => package,
+            None => {
+                error!(msg: "Package '{}' doesn't exist", self.package);
```
BraveChicken1:
```suggestion
                error!(msg: "Package '{}' is not installed", self.package);
```

src/cli/commands/info.rs
```diff
@@ -0,0 +1,148 @@
...
+
+        for dependent in &package_version.dependents {
+            print!("\n  - {dependent}");
+        }
+        println!();
+
+        // Early return if verbose is disabled
+        if !self.verbose {
+            return;
+        }
```
BraveChicken1: I would also only include dependent in the verbose mode

src/installer/installer.rs
```diff
@@ -365,28 +311,21 @@ impl<'a> Installer<'a> {
...
+            && !self.register.is_dependency(optional_id)
+            && *node.get_label() == DependencyTypes::Build
+        {
             self.uninstall(optional_id)?;
         }
 
+        // Remove children
+        for child in node.get_children_mut() {
+            self.remove_build_dependencies(child)?;
+        }
```
BraveChicken1, line 334: It would maybe be better to change this into an iterative method

src/utils/tree.rs
```diff
@@ -0,0 +1,201 @@
...
+    repositories::{error::RepositoryError, manager::RepositoryManager},
+    storage::package_register::PackageRegister,
+};
+
+#[derive(Debug)]
+pub struct Node<V, T: Eq> {
+    id: PackageId,
+    value: V,
+    children: Vec<Node<V, T>>,
+    label: T,
```
BraveChicken1: Since the variable is called `label`, it would probably be neater to use `L` for the generic type

src/utils/tree.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    pub fn get_id(&self) -> &PackageId {
+        &self.id
+    }
+
+    pub fn get_value(&self) -> &V {
+        &self.value
+    }
+
+    pub fn get_children_as_id(&self, group: Option<T>) -> HashSet<PackageId> {
```
```suggestion
    pub fn get_children_ids(&self, group: Option<T>) -> HashSet<PackageId> {
```

BraveChicken1: Does group specify the label to filter on? Shouldn't it then be called `label`?

src/utils/tree.rs
```diff
@@ -0,0 +1,201 @@
...
+    pub fn new(package_id: &PackageId, register: &PackageRegister) -> Result<Self, TreeError> {
+        Node::new_impl(package_id, register, &|_| ((), ()))
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub enum DependencyTypes {
+    Normal,
+    Build,
+}
```
BraveChicken1: This enum should maybe be defined in in the installer module, since it belongs to the install

src/utils/tree.rs
```diff
@@ -0,0 +1,201 @@
...
+            dependencies
+                .into_iter()
+                .map(|d| {
+                    let install_meta = InstallMeta::new(manager, d)?;
+                    let latest_version = install_meta.package_metadata.get_latest_version(TARGET_ARCHITECTURE)?;
+                    let dependency_id = PackageId::new(d.get_name(), latest_version).expect("Expected valid dependency");
+                    Self::new_from_meta_impl(&dependency_id, install_meta, manager, DependencyTypes::Normal, include_build)
+                })
+                .collect::<Result<_, _>>()?
+        };
```
BraveChicken1: I would extract the dependencies.map and then conditionally add the build dependencies

BraveChicken1 (changes requested):
Looks good overall!

src/installer/installer.rs
Max13245: Yes maybe, but that might be a change for when we implement the parallelization.

src/utils/tree.rs
Max13245: Yes, I mainly named it group so we don't have label as a variable two times, but I could change that.

src/utils/tree.rs
Max13245: I was planning on putting the duplicate code into a separate method instead.

src/cli/commands/info.rs
Max13245: Do you mean to only add dependents (that makes sense) or only dependents as verbose data?

BraveChicken1: What I mean is that we move the dependent output into the verbose mode

src/utils/tree.rs
BraveChicken1: That would also be fine

BraveChicken1 (approved):

PR #40, opened 2026-03-03
Implement os version specific targets

BraveChicken1:
This PR implement support for os version specific targets in package metadata.

### DIFF
--- a/src/installer/types/version_bounds.rs
+++ b/src/installer/types/version_bounds.rs
@@ -62,6 +62,18 @@ impl VersionBounds {
 
         Ok(bounds)
     }
+
+    pub fn covers(&self, version: &Version) -> bool {
+        match self {
+            VersionBounds::Range(lower, upper) if lower <= version && upper >= version => true,
+            VersionBounds::Lower(lower) if version < lower => true,
+            VersionBounds::LowerEqual(lower) if version <= lower => true,
+            VersionBounds::Higher(higher) if version > higher => true,
+            VersionBounds::HigherEqual(higher) if version >= higher => true,
+            VersionBounds::Equal(equal) if version == equal => true,
+            _ => false,
+        }
+    }
 }
 
 #[cfg(test)]
--- a/src/installer/types/version_bounds.rs
+++ b/src/installer/types/version_bounds.rs
@@ -2,7 +2,7 @@ use std::str::FromStr;
 
 use crate::installer::types::{DependencyParserError, Version};
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
 pub enum VersionBounds {
     Range(Version, Version),
     Lower(Version),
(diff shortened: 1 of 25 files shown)


src/installer/types/version_bounds.rs
```diff
@@ -62,6 +62,18 @@ impl VersionBounds {
...
+        match self {
+            VersionBounds::Range(lower, upper) if lower <= version && upper >= version => true,
+            VersionBounds::Lower(lower) if version < lower => true,
+            VersionBounds::LowerEqual(lower) if version <= lower => true,
+            VersionBounds::Higher(higher) if version > higher => true,
+            VersionBounds::HigherEqual(higher) if version >= higher => true,
+            VersionBounds::Equal(equal) if version == equal => true,
+            _ => false,
+        }
+    }
```
Max13245, line 76: Nice :)

src/platforms/os.rs
```diff
@@ -0,0 +1,210 @@
...
+    Unknown,
+}
+
+impl Os {
+    pub fn is_unix(&self) -> bool {
+        match self {
+            Self::MacOs | Self::Linux => true,
+            _ => false,
+        }
+    }
```
Max13245, line 22: You could change this to the following, I'm not sure if this is  cleaner though. The match statement is also okay.
```
*self == Self::MacOs || *self == Self::Linux
```

src/platforms/os.rs
```diff
@@ -0,0 +1,210 @@
...
+    },
+    Unknown,
+}
+
+static CURRENT_VERSION: LazyLock<OsVersion> = LazyLock::new(|| match OsVersion::get_version() {
+    Some(value) => value,
+    None => OsVersion::Unknown,
+});
+
+impl OsVersion {
```
Max13245, line 47: For getting OS versions and in case of Linux also distros. There is a crate which seems to do just that, however it is not actively maintained. The repo looks reasonable though (and the dev seems to know what he is doing), so I thought we should at least have a look:
crate: https://crates.io/crates/os-version
github: https://github.com/kpcyrd/os-version

Max13245: Also, this reads files instead of executing commands, which is probably more reliable.

src/repositories/types/target_bounds.rs
```diff
@@ -0,0 +1,224 @@
...
+        // Split addition from name
+        let (name, addition) = match name.split_once(':') {
+            Some((name, addition)) => (name, Some(addition)),
+            None => (name, None),
+        };
+
+        let name = TargetName::from_str(name)?;
+
+        // Check if additions are allowed
+        if addition.is_some() {
```
Max13245: Can you please change this to the following:
```
if addition.is_some() && !matches!(name.get_os(), Some(Os::Linux)) {
            return Err(TargetBoundsError::AdditionNotAllowed);
        }
```

Max13245 (commented):
Overall looks good!

src/platforms/os.rs
BraveChicken1: I refactored the Linux get_version to only read files now. For macOS it would also be possible, but it would require introducing new dependencies, while the current approach seems easier.

Max13245: I would say it's okay for now, but we should look at this in the future.

Max13245 (approved):

PR #41, opened 2026-03-03
Add update command

Max13245:
This PR adds the update command.

### DIFF
--- a/src/cli/commands/update.rs
+++ b/src/cli/commands/update.rs
@@ -1,0 +1,36 @@
+use clap::Args;
+
+use crate::{
+    cli::commands::HandleCommand,
+    config::Config,
+    installer::{Installer, InstallerOptions, types::OptionalPackageId},
+    platforms::TARGET_ARCHITECTURE,
+    repositories::manager::RepositoryManager,
+    storage::package_register::PackageRegister,
+    utils::unwrap_or_exit::UnwrapOrExit,
+};
+
+// TODO: Add a version parameter to specify which version to update or downgrade to
+#[derive(Args, Debug)]
+pub struct UpdateArgs {
+    /// The name of the packages to update, with an optional version specified with NAME@VERSION
+    optional_id: OptionalPackageId,
+}
+
+impl HandleCommand for UpdateArgs {
+    fn handle(&self, config: &Config, manager: &RepositoryManager) {
+        let register_dir = PackageRegister::get_default_path(config);
+        let mut register = PackageRegister::from(&register_dir).unwrap_or_exit(1);
+
+        let (_, package_meta) = manager.read_package(&self.optional_id.name).unwrap_or_exit(1);
+        let latest_version = package_meta.get_latest_version(TARGET_ARCHITECTURE).unwrap_or_exit(1);
+
+        let options = InstallerOptions::default().skip_active(true).skip_symlinking(true);
+        let mut installer = Installer::new(config, &mut register, manager, options);
+
+        installer.update(&self.optional_id, latest_version).unwrap_or_exit(1);
+
+        // Save changes
+        register.save_to(&register_dir).unwrap_or_exit(1);
+    }
+}
--- a/src/cli/commands/update.rs
+++ b/src/cli/commands/update.rs
@@ -1,9 +1,14 @@
+use std::process::exit;
+
 use clap::Args;
 
 use crate::{
-    cli::commands::HandleCommand,
+    cli::{commands::HandleCommand, display::logging::error},
     config::Config,
-    installer::{Installer, InstallerOptions, types::OptionalPackageId},
+    installer::{
+        Installer, InstallerOptions,
+        types::{OptionalPackageId, Version},
+    },
     platforms::TARGET_ARCHITECTURE,
     repositories::manager::RepositoryManager,
     storage::package_register::PackageRegister,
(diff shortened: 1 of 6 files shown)


src/cli/commands/update.rs
```diff
@@ -0,0 +1,54 @@
...
+    utils::unwrap_or_exit::UnwrapOrExit,
+};
+
+// TODO: Add a version parameter to specify which version to update or downgrade to
+#[derive(Args, Debug)]
+pub struct UpdateArgs {
+    /// The name of the packages to update, with an optional version specified with NAME@VERSION
+    optional_id: OptionalPackageId,
+
+    /// A new version to update to. Can only be a high version then the current version
```
```suggestion
    /// The version to update to. This can only be a higher version than the current version
```

src/cli/commands/update.rs
```diff
@@ -0,0 +1,54 @@
...
+        Installer, InstallerOptions,
+        types::{OptionalPackageId, Version},
+    },
+    platforms::TARGET_ARCHITECTURE,
+    repositories::manager::RepositoryManager,
+    storage::package_register::PackageRegister,
+    utils::unwrap_or_exit::UnwrapOrExit,
+};
+
+// TODO: Add a version parameter to specify which version to update or downgrade to
```
BraveChicken1: Isn't this the `new_version` field?

src/cli/commands/update.rs
```diff
@@ -0,0 +1,54 @@
...
+#[derive(Args, Debug)]
+pub struct UpdateArgs {
+    /// The name of the packages to update, with an optional version specified with NAME@VERSION
```
```suggestion
    /// The name of the package to update, with an optional version specified with NAME@VERSION
```

src/cli/commands/update.rs
```diff
@@ -0,0 +1,54 @@
...
+    fn handle(&self, config: &Config, manager: &RepositoryManager) {
+        let register_dir = PackageRegister::get_default_path(config);
+        let mut register = PackageRegister::from(&register_dir).unwrap_or_exit(1);
+
+        let (_, package_meta) = manager.read_package(&self.optional_id.name).unwrap_or_exit(1);
+
+        let new_version = match &self.new_version {
+            Some(version) => version,
+            None => package_meta.get_latest_version(TARGET_ARCHITECTURE).unwrap_or_exit(1),
+        };
```
BraveChicken1, line 37: Shouldn't we use the same repository as the one we installed the package from originally?

src/installer/error.rs
```diff
@@ -27,12 +32,30 @@ pub enum InstallerError {
...
 
-    #[error("Package '{package_name}' with version '{}' does not exist.", version.clone().unwrap_or("any".to_string()))]
+    #[error("Package '{package_name}' with version '{}' is not installed.", version.clone().unwrap_or("any".to_string()))]
     PackageNotFound {
         package_name: String,
         version: Option<String>,
     },
 
+    #[error("Package '{package_id}' is already installed.")]
+    ExistError {
```
BraveChicken1:
```suggestion
    AlreadyExistError {
```

src/installer/error.rs
```diff
@@ -27,12 +32,30 @@ pub enum InstallerError {
...
+    #[error("Multiple versions are installed, so only package name input is not specific enough. Please provide a version as well.")]
+    SpecificityError,
+
+    #[error("The new version '{new_version}' of this package cannot satisfy all dependents from the old package version.")]
+    SatisfyError {
+        new_version: Version,
+    },
+
+    #[error("Could not update, the given version '{new_version}' is lower then the current version.")]
+    UpdateVersionError {
```
BraveChicken1:
```suggestion
    VersionTooLowError {
```

src/installer/error.rs
```diff
@@ -27,12 +32,30 @@ pub enum InstallerError {
...
+    #[error("Package '{package_id}' is already installed.")]
+    ExistError {
+        package_id: PackageId,
+    },
+
+    #[error("Multiple versions are installed, so only package name input is not specific enough. Please provide a version as well.")]
+    SpecificityError,
+
+    #[error("The new version '{new_version}' of this package cannot satisfy all dependents from the old package version.")]
+    SatisfyError {
```
BraveChicken1, line 50:
```suggestion
    NotAllSatisfiedError {
```

BraveChicken1 (changes requested):

src/cli/commands/update.rs
Max13245: I would say that that's not necessary, otherwise you can't update if the newer version is not in the original repo. A user can already define de repo's they want to use in their config. If we want something more precise, I think we should add a flag to the install and update command.

src/installer/error.rs
Max13245: I changed it to AlreadyInstalledError instead.

src/installer/error.rs
Max13245: I think SatisfyError is good enough, this would not add that much more clarity.

src/cli/commands/update.rs
BraveChicken1: I agree, however I'm not fully sure if it could result in some strange mismatch when the updated version is from another repository.

src/installer/error.rs
BraveChicken1: SatisfyError sounds a bit vague, but I also can't really think of a better name for it.

Max13245: I think an error can be a bit vague, as long as the error message is clear.

src/cli/commands/update.rs
Max13245: I don't think it will, because the update will be the same as uninstalling and then installing from another repo. The only thing that I can think of is that if the symlinked state defined in one repo is different from the definition in another repo. But this would be rare and one of the repo's definitions would be wrongly implemented.

BraveChicken1 (approved):

PR #42, opened 2026-03-07
Fix dependency already satisfied bug

Max13245:
Fixes the dependency bug where an already satisfied dependency was not detected when a later version is available.


src/utils/tree.rs
```diff
@@ -153,38 +154,65 @@ impl Node<InstallMeta, DependencyTypes> {
...
-            self.children.push(Self::new_from_dependency(manager, dependency, DependencyTypes::Build, false)?);
+    /// Note that this only applies for the build dependencies of the current node, the
+    /// dependencies of the current node will be satisfied with pre-builds.
+    pub fn expand_node_with_build(&mut self, manager: &RepositoryManager, register: &PackageRegister) -> Result<(), TreeError> {
+        let value = match &self.value {
+            Some(value) => value,
+            None => {
+                // Should be unreachable
+                error!(msg: "Cannot expand a node without a value.");
+                exit(1);
```
BraveChicken1: This should probably not exit the application, but return an error.

src/utils/tree.rs
```diff
@@ -193,14 +221,29 @@ impl Node<InstallMeta, DependencyTypes> {
...
+                id: package.package_id.clone(),
+                value: None,
+                children: Vec::new(),
+                label,
+            });
+        }
+
+        // Use the latest version if the dependency is not yet satisfied
+        let (repository_id, package_metadata) = manager.read_package(dependency.get_name())?;
+        dbg!(&repository_id);
```
BraveChicken1: If you want to keep a debug print here, please use the `debug!` macro.

BraveChicken1 (changes requested):
Looks good overall!

Do you want to split out the InstallMeta things from the tree util in this PR, or in a different one?

src/utils/tree.rs
Max13245: I left this by excedent, I'll remove it thnx.

src/utils/tree.rs
Max13245: Alright, although I do think we error and exit in some other places where code should be unreachable.

Max13245:
I think moving the install tree implementation could better be done on another branch.

BraveChicken1 (approved):

PR #43, opened 2026-03-07
Refactor version bounds

Max13245:
This refactors the version bounds. VersionBounds are now stored in a VersionInterval struct. 
It also fixes the bug where an error is thrown if the latest satisfying version for a dependency is not supported by the current target, even though a lower supported version would also satisfy the dependency.
To fix this the package metadata now doesn't contain the latest version, but the supported versions for each target.

### DIFF
--- a/src/installer/types/dependency.rs
+++ b/src/installer/types/dependency.rs
@@ -14,7 +14,6 @@ pub enum DependencyParserError {
     InvalidDependencyName,
 }
 
-// TODO: Dependency names should be validated like in the PackageId also implement a method which takes a version and then returns a PackageId
 #[derive(Debug, Clone)]
 pub struct Dependency {
     name: String,
@@ -95,23 +94,17 @@ impl Dependency {
         &self.name
     }
 
-    // TODO: Should only except package_id, version shouldn't be optional
-    pub fn satisfied(&self, name: &str, version: Option<&Version>) -> bool {
+    pub fn satisfied(&self, name: &str, version: &Version) -> bool {
         if self.name != name {
             return false;
         }
-
-        let version = match version {
-            Some(version) => version,
-            None => return true,
-        };
 
         if self.version_ranges.is_empty() {
             return true;
         }
 
         for range in &self.version_ranges {
-            if range.covers(version) {
+            if range.covers(&version) {
                 return true;
             }
         }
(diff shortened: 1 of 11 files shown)


Max13245:
A couple of doubts:
- It is necessary to know the supported versions of each target. However, if it's defined by version intervals we need to make sure these intervals don't contain any greater (or equals) than bounds, because this would be impossible. Do we not want to use the intervals here or should there be a check in the deserialize (currently the check is present, but not in the deserialize)?

- Should we also use the version intervals for the package versions (assuming we use them for the target specific as well)? This would also require the get_latest_dependency_version method to be changed. I already have a solution in mind. This solutions implementation is some what ugly, although technically a tiny bit more efficient than what we have now.

EDIT: Already resolved

src/installer/types/dependency.rs
```diff
@@ -95,28 +93,12 @@ impl Dependency {
         &self.name
     }
 
-    // TODO: Should only except package_id, version shouldn't be optional
-    pub fn satisfied(&self, name: &str, version: Option<&Version>) -> bool {
+    pub fn satisfied(&self, name: &str, version: &Version) -> bool {
```
BraveChicken1, line 96: Is there a reason this accepts a name and version, and not a PackageId?

src/installer/types/version_intervals.rs
```diff
@@ -0,0 +1,170 @@
...
+        D: serde::Deserializer<'de>,
+    {
+        let string: String = de::Deserialize::deserialize(deserializer)?;
+
+        Ok(Self::from_str_intervals(&string).map_err(de::Error::custom)?)
+    }
+}
+
+impl VersionIntervals {
+    pub fn from_str_intervals(intervals: &str) -> Result<Self, VersionError> {
```
BraveChicken1: This could be an implementation of the FromStr trait

src/installer/types/version_intervals.rs
```diff
@@ -0,0 +1,170 @@
...
+        if !Self::validate_intervals(&version_bounds) {
+            return Err(VersionError::InvalidInterval);
+        }
+
+        Ok(Self { version_bounds })
+    }
+
+    /// Checks if the intervals are valid, the intervals are valid if they don't overlap and are in order.
+    /// True is returned if the intervals are valid, otherwise false.
+    fn validate_intervals(version_bounds: &Vec<VersionBounds>) -> bool {
```
BraveChicken1:
```suggestion
    fn bounds_valid(version_bounds: &Vec<VersionBounds>) -> bool {
```

src/installer/types/version_intervals.rs
```diff
@@ -0,0 +1,170 @@
...
+                None => {
+                    previous = Some(&bound);
+                    continue;
+                },
+            };
+
+            let low_version = match bound {
+                VersionBounds::Range(low, _) => low,
+                VersionBounds::Lower(_) => return false,
+                VersionBounds::LowerEqual(_) => return false,
```
BraveChicken1, line 70: Doesn't this cause VersionBounds Lower and LowerEqual to always be invalid?

src/installer/types/version_intervals.rs
```diff
@@ -0,0 +1,170 @@
...
+                VersionBounds::Range(_, high) if *low_version <= *high => return false,
+                VersionBounds::Lower(version) if low_version < version => return false,
+                VersionBounds::LowerEqual(version) if low_version <= version => return false,
+                VersionBounds::Higher(_) => return false,
+                VersionBounds::HigherEqual(_) => return false,
+                VersionBounds::Equal(version) if *low_version <= *version => return false,
+                _ => {},
+            }
+
+            previous = Some(bound)
```
BraveChicken1, line 88: Shouldn't this check each bound with each other bound, to check if one of the bounds in the list overlaps with another?

src/installer/types/version_intervals.rs
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        true
+    }
+
+    pub fn covers(&self, version: &Version) -> bool {
+        // If version bounds are empty, version satisfies the bounds
+        if self.version_bounds.is_empty() {
+            return true;
+        }
```
BraveChicken1, line 98: Is this wanted behaviour in every scenario?

src/repositories/types/package.rs
```diff
@@ -12,41 +12,60 @@ use crate::{
...
+            .latest_version_impl(|v| !dependency.satisfied(&self.name, v) || !supported.covers(v))?
+            .ok_or(RepositoryError::SupportError(dependency.to_string()))?)
+    }
+
+    /// A generic method to get the latest version of the current package.
+    /// If any of the checks are true for the current version we continue for that version.
+    fn latest_version_impl<F>(&self, checks: F) -> Result<Option<&Version>>
+    where
+        F: Fn(&Version) -> bool,
+    {
```
BraveChicken1: In my opinion this generic method adds unnecessary complexity to this code, I think it would be better to have some duplicate code here instead.

src/repositories/types/target_bounds.rs
```diff
@@ -154,38 +154,26 @@ impl TargetBounds {
...
-            }
-        }
-
-        false
+        self.version_intervals.covers(version)
     }
 
     fn calculate_priority(&self) -> u32 {
-        if self.addition.is_none() && self.version_bounds.is_empty() {
+        if self.addition.is_none() && self.version_intervals.get_version_bounds().is_empty() {
```
BraveChicken1: Maybe VersionIntervals could have a method `is_empty`, since severals parts seem to need this

BraveChicken1 (changes requested):
Looks good overall!

Don't forget to also change the core repository according to the changes here

src/installer/types/dependency.rs
Max13245: The reason I did this is because in the get_latest_dependency_version method inside package.rs we call this method. We would have to create a package id in that case, but then we have to deal with the error.

What I could do is that I fix that todo for the PackageMeta struct (validate the package name). Then I can create a method to_package_id on the PackageMeta struct which can expect the PackageId::new() method (like in some other places).

src/installer/types/version_intervals.rs
Max13245: Yes, good idea

src/installer/types/version_intervals.rs
Max13245: Yes, I would say that if there are no bounds the version is not bound by anything, so any version is covered.

src/repositories/types/package.rs
Max13245: Yes I was not sure about this, I'll change it.

src/installer/types/version_intervals.rs
Max13245: No, because the first iteration never gets there, if there is a second iteration it should always return false. Anything before it would be overlapping with the lower (or equal) bound or the bounds are not in order. I see how this can be a bit confusing, I'll add a comment.

src/installer/types/version_intervals.rs
Max13245: No, that's the beauty of this. We don't have to compare everything with each other, because the intervals need to be in order.

src/installer/types/dependency.rs
BraveChicken1: I'm not sure if the PackageMeta should have a to_package_id method. Just keep it like this for now, we'll look into this later.

Max13245: Oki, should I already implement the check for the PackageMeta name though?

BraveChicken1 (approved):

PR #47, opened 2026-03-11
Bugfix for removal of build dependencies which were previously installed

Max13245:
This PR includes a bugfix for removal of build dependencies which were installed before installation. It now keeps track of the newly installed packages and only removes those (if they are not also a new dependency).

### DIFF
--- a/src/installer/installer.rs
+++ b/src/installer/installer.rs
@@ -135,30 +135,34 @@ impl<'a> Installer<'a> {
             target_bounds,
         };
 
-        let mut newly_installed = Vec::new();
+        let mut installed_build_deps = Vec::new();
         match self.options.build_source {
             true => {
                 let dependency_tree = Node::new_from_meta_build(&package_id, root_meta, self.repository_manager, self.register)?;
-                self.install_nodes_build(&dependency_tree, &mut newly_installed)?;
+                self.install_nodes_build(&dependency_tree, &mut installed_build_deps)?;
             },
             false => {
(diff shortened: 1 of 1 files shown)


src/installer/installer.rs
```diff
@@ -135,31 +135,30 @@ impl<'a> Installer<'a> {
             target_bounds,
         };
 
-        let mut tree = match self.options.build_source {
+        let mut newly_installed = Vec::new();
```
```suggestion
        let mut installed_build_deps = Vec::new();
```

src/installer/installer.rs
```diff
@@ -135,31 +135,30 @@ impl<'a> Installer<'a> {
...
         if !self.options.keep_build {
-            self.remove_build_dependencies(&mut tree)?;
+            self.remove_build_dependencies(&newly_installed)?;
         }
 
         Ok(())
     }
 
-    fn install_nodes(&mut self, node: &mut Node<Option<InstallMeta>, DependencyTypes>) -> Result<()> {
+    fn install_nodes(&mut self, node: &mut Node<Option<InstallMeta>, DependencyTypes>, installed: &mut Vec<PackageId>) -> Result<()> {
```
```suggestion
    fn install_nodes(&mut self, node: &mut Node<Option<InstallMeta>, DependencyTypes>, installed_build_deps: &mut Vec<PackageId>) -> Result<()> {
```

BraveChicken1 (changes requested):

BraveChicken1 (approved):

PR #49, opened 2026-03-21
Complete refactor of the tree struct

Max13245:
- There is now a tree builder which builds the tree with help of a generic expander and populator.
- The logic for the install tree is now separated from the generic tree struct in the util.
- There are now two build flags, --build and --build-all (to also build dependencies from source).

### DIFF
--- a/src/cli/commands/install.rs
+++ b/src/cli/commands/install.rs
@@ -3,7 +3,7 @@ use clap::Args;
 use crate::{
     cli::{commands::HandleCommand, display::logging::error},
     config::Config,
-    installer::{InstallTypes, Installer, InstallerOptions, types::OptionalPackageId},
+    installer::{InstallLabel, InstallTypes, Installer, InstallerOptions, types::OptionalPackageId},
     repositories::manager::RepositoryManager,
     storage::package_register::PackageRegister,
     utils::unwrap_or_exit::UnwrapOrExit,
@@ -45,15 +45,17 @@ impl HandleCommand for InstallArgs {
 
         // Determine the install type. Note that clap already check if build and build-all are both set (which should not be possible).
         let install_type = if self.build {
-            InstallTypes::Build { is_dependency: false }
+            InstallTypes::Build
         } else if self.build_all {
-            InstallTypes::BuildAll { is_dependency: false }
+            InstallTypes::BuildAll
         } else {
-            InstallTypes::Prebuild { is_dependency: false }
+            InstallTypes::Prebuild
         };
 
+        let install_label = InstallLabel::new(install_type, false);
+
         let installer_options = InstallerOptions::default()
-            .install_type(install_type)
+            .install_label(install_label)
             .skip_symlinking(self.skip_symlinking)
             .skip_active(self.skip_active)
             .keep_build(self.keep_build);
--- a/src/cli/commands/install.rs
+++ b/src/cli/commands/install.rs
@@ -3,7 +3,7 @@ use clap::Args;
 use crate::{
     cli::{commands::HandleCommand, display::logging::error},
     config::Config,
-    installer::{Installer, InstallerOptions, types::OptionalPackageId},
+    installer::{InstallTypes, Installer, InstallerOptions, types::OptionalPackageId},
     repositories::manager::RepositoryManager,
     storage::package_register::PackageRegister,
     utils::unwrap_or_exit::UnwrapOrExit,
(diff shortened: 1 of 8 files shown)


src/cli/commands/install.rs
```diff
@@ -39,8 +43,19 @@ impl HandleCommand for InstallArgs {
...
+            InstallTypes::BuildAll
+        } else {
+            InstallTypes::Prebuild
+        };
+
+        let install_label = InstallLabel::new(install_type, false);
+
         let installer_options = InstallerOptions::default()
-            .build_source(self.build)
+            .install_label(install_label)
```
BraveChicken1: I think it would be neater to use just InstallTypes in the InstallerOptions, the installer will then create an InstallLabel from this.

src/installer/install_tree.rs
```diff
@@ -0,0 +1,149 @@
...
+        manager::RepositoryManager,
+        types::{PackageMeta, PackageVersionMeta, TargetBounds},
+    },
+    storage::package_register::PackageRegister,
+    utils::tree::{self, Node},
+};
+
+/// Represents the different types of installing a package.
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub enum InstallTypes {
```
BraveChicken1:
```suggestion
pub enum InstallType {
```

src/installer/install_tree.rs
```diff
@@ -0,0 +1,149 @@
...
+    /// Expands an install tree after it has been build. The current node will be changed to a build node
+    /// and its children will be updated accordingly.
+    pub fn expand_with_build(&mut self, register: &PackageRegister, manager: &RepositoryManager) -> tree::Result<()> {
+        self.set_label(InstallLabel {
+            install_type: InstallTypes::Build,
+            is_dependency: self.get_label().is_dependency(),
+        });
+
+        // Clear all the children to avoid duplicates
+        self.get_children_mut().clear();
```
BraveChicken1: Doesn't this cause the subtree to be fully rebuild? Do we really want that behaviour?

src/installer/installer.rs
```diff
@@ -135,34 +104,28 @@ impl<'a> Installer<'a> {
...
+            .root(package_id, Some(root_meta), self.options.install_label.clone())
+            .expander(InstallNode::expander)
+            .populator(|(d, l)| InstallNode::populator(self.register, self.repository_manager, &d, l))
+            .build()?;
+
+        self.install_nodes(&mut dependency_tree)?;
 
         if !self.options.keep_build {
-            self.remove_build_dependencies(&installed_build_deps)?;
+            self.remove_build_dependencies(&dependency_tree, &dependency_tree)?;
```
BraveChicken1: It seems that this function takes the same argument twice.

src/installer/installer.rs
```diff
@@ -171,11 +134,19 @@ impl<'a> Installer<'a> {
             None => return Ok(()),
         };
 
+        let dependencies = node.get_children_ids(Some(InstallLabel::is_dependency));
+
+        // Check if the current package should be build from source
+        if matches!(node.get_label().get_type(), InstallTypes::Build) || matches!(node.get_label().get_type(), InstallTypes::BuildAll) {
```
```suggestion
        if matches!(node.get_label().get_type(), InstallTypes::Build | InstallTypes::BuildAll) {
```

src/installer/installer.rs
```diff
@@ -364,20 +309,29 @@ impl<'a> Installer<'a> {
...
     }
 
-    fn remove_build_dependencies(&mut self, installed: &Vec<PackageId>) -> Result<()> {
-        // Loop through the installed build dependencies in reverse, because then the dependents
-        // are uninstalled before the dependencies due to the install order.
-        for package_id in installed.iter().rev() {
-            let optional_id = &OptionalPackageId::from(package_id.clone());
+    /// Removes the build dependencies recursively. There are early returns to make sure that the
+    /// package is not removed if it was already installed, is not installed anymore or is a dependency.
+    fn remove_build_dependencies(&mut self, parent: &InstallNode, root: &InstallNode) -> Result<()> {
```
BraveChicken1: Couldn't the second argument be just a boolean? `is_root: bool`

src/utils/tree.rs
```diff
@@ -127,122 +184,32 @@ impl<V, L: Eq> Node<V, L> {
...
-                DependencyTypes::Build,
-                false,
-            )?);
-        }
+pub type EmptyNode = Node<(), ()>;
 
-        Ok(())
+/// An empty node implementation (node without values or labels).
+impl EmptyNode {
+    pub fn build_simple(package_id: PackageId, register: &PackageRegister) -> Result<EmptyNode> {
```
```suggestion
    pub fn build_simple_tree(package_id: PackageId, register: &PackageRegister) -> Result<EmptyNode> {
```

BraveChicken1 (changes requested):
Looks good overall!

src/installer/installer.rs
Max13245: Yes, good idea, I'll change that.

src/installer/install_tree.rs
Max13245: Yes it does, there is only one other way of fixing this. We would have to check if the child already exists in the expander method before adding it. We can't make the children Vec a HashSet, because the nodes need to be mutable in some cases (specifically if we want to expand the tree after initial creation). So every time we add a node we would have to loop over all existing children.

Maybe another way could be to only add the build dependencies. This would be difficult to do in the expander method though, so then we would probably need a separate method which only adds build dependencies. Unless we do an ugly trick where we check if a node already has children and then only add the build dependencies, but this would be very unreliable.

BraveChicken1 (approved):

PR #51, opened 2026-03-25
Prettier (error) logging

Max13245:
Adds several styling features:
- Print packages alphabetically in the list command
- Add debug errors, so there aren't too many (unnecessary) errors.
- Create more detailed logging in the install and uninstall methods.
- Add success message to link, unlink, update, install and uninstall commands.
- A verbose mode for the build, so build logging is only done with it enabled. It now has a spinner.

Small bugfix:
- An (un)install with multiple packages should have at least one package argument, this is now enforced.

Small feature:
- Duplicates are now not allowed anymore in the (un)install package arguments.

### DIFF
--- a/src/cli/commands/install.rs
+++ b/src/cli/commands/install.rs
@@ -49,6 +49,7 @@ impl HandleCommand for InstallArgs {
             let mut duplicate_string = String::new();
             for duplicate in duplicates {
                 duplicate_string.push_str(&duplicate.to_string());
+                duplicate_string.push(' ');
             }
 
             error!(msg: "Duplicate package arguments are not allowed. The following duplicates were found: {duplicate_string}");
@@ -80,8 +81,9 @@ impl HandleCommand for InstallArgs {
 
         // Install all packages
         for package_id in &self.packages {
-            if let Err(error) = installer.install(&package_id) {
-                error!(error, "Cannot install package {package_id}");
+            match installer.install(&package_id) {
+                Ok(installed_package) => println!("Successfully installed {installed_package}"),
+                Err(error) => error!(error, "Cannot install package {package_id}"),
             }
         }
 
--- a/src/cli/commands/install.rs
+++ b/src/cli/commands/install.rs
@@ -1,3 +1,5 @@
+use std::process::exit;
+
 use clap::Args;
 
 use crate::{
@@ -6,7 +8,7 @@ use crate::{
     installer::{InstallType, Installer, InstallerOptions, types::OptionalPackageId},
     repositories::manager::RepositoryManager,
     storage::package_register::PackageRegister,
-    utils::unwrap_or_exit::UnwrapOrExit,
+    utils::{duplicates, unwrap_or_exit::UnwrapOrExit},
 };
 
 /// Installs the specified packages, if a version is given that version will be installed,
@@ -41,6 +43,18 @@ pub struct InstallArgs {
 
 impl HandleCommand for InstallArgs {
     fn handle(&self) {
+        // Check for duplicates, because installing twice will result in a confusing error
+        let duplicates = duplicates::get_duplicates(&self.packages);
+        if !duplicates.is_empty() {
+            let mut duplicate_string = String::new();
+            for duplicate in duplicates {
+                duplicate_string.push_str(&duplicate.to_string());
+            }
+
+            error!(msg: "Duplicate package arguments are not allowed. The following duplicates were found: {duplicate_string}");
+            exit(1);
+        }
+
         let config = Config::from(&Config::get_default_path()).unwrap_or_exit_msg("Cannot load config", 1);
         let manager = RepositoryManager::new(&config);
         let register_dir = PackageRegister::get_default_path(&config);
@@ -63,7 +77,6 @@ impl HandleCommand for InstallArgs {
         let mut installer = Installer::new(&config, &mut register, &manager, installer_options);
 
         // TODO: Check if this exists as an external package (possibly leading to conflicts) (if so, add to external packages)
-        // TODO: check for duplicate packages in Vec
 
         // Install all packages
         for package_id in &self.packages {
--- a/src/cli/commands/install.rs
+++ b/src/cli/commands/install.rs
@@ -15,7 +15,7 @@ use crate::{
 #[derive(Args, Debug)]
 pub struct InstallArgs {
     /// The name of the packages to install, with an optional version specified with NAME@VERSION
-    #[arg(num_args(0..))]
+    #[arg(required = true)]
     pub packages: Vec<OptionalPackageId>,
 
     /// True to build from source locally, false to use a prebuild version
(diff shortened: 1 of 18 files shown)


src/cli/commands/install.rs
```diff
@@ -6,13 +8,13 @@ use crate::{
...
     storage::package_register::PackageRegister,
-    utils::unwrap_or_exit::UnwrapOrExit,
+    utils::{duplicates, unwrap_or_exit::UnwrapOrExit},
 };
 
 #[derive(Args, Debug)]
 pub struct InstallArgs {
     /// The name of the packages to install, with an optional version specified with NAME@VERSION
-    #[arg(num_args(0..))]
+    #[arg(required = true)]
```
BraveChicken1, line 20:
```suggestion
    #[arg(num_args(1..))]
```

src/cli/commands/uninstall.rs
```diff
@@ -6,18 +8,31 @@ use crate::{
...
-    utils::unwrap_or_exit::UnwrapOrExit,
+    utils::{duplicates, unwrap_or_exit::UnwrapOrExit},
 };
 
 #[derive(Args, Debug)]
 pub struct UninstallArgs {
-    /// The name of the packages to install, with an optional version specified with NAME@VERSION
-    #[arg(num_args(0..))]
+    /// The names of the packages to install, with an optional version specified with `<name>[@version]`.
+    #[arg(required = true)]
```
BraveChicken1, line 20:
```suggestion
    #[arg(num_args(1..))]
```

src/cli/display/reader.rs
```diff
@@ -11,10 +11,10 @@ pub struct ReaderWithProgress<R: Read + Seek> {
 
 impl<R: Read + Seek> ReaderWithProgress<R> {
     /// Creates a new reader with a progress bar.
-    pub fn new(reader: R, size: u64) -> Self {
+    pub fn new(reader: R, size: u64, prefix: String) -> Self {
```
BraveChicken1:
```suggestion
    pub fn new(reader: R, size: u64, bar_prefix: String) -> Self {
```

src/installer/installer.rs
```diff
@@ -200,7 +205,7 @@ impl<'a> Installer<'a> {
             scripts::download_script(self.repository_manager, &script_path, &package_id.name, &install_meta.repository_id)?;
         if let Some(script_file) = downloaded_script {
             let script_data = ScriptData::new(&script_file, &install_directory, &version_meta.version, self.config, &script_args);
-            scripts::run_pre_script(&script_data, &install_directory)?;
+            scripts::run_pre_script(&script_data, &install_directory, self.options.verbose)?;
```
BraveChicken1: We probably want some output from preinstall, postinstall and uninstall scripts. These could be used to perform tasks that need user input or notify the user about things. We would then need the output. We could create a `PACKIT_VERBOSE` environment variable to allow the script to filter the output correctly.

src/installer/installer.rs
```diff
@@ -433,28 +447,34 @@ impl<'a> Installer<'a> {
...
 
             if let Some(newest) = other_versions.last() {
+                println!("Set active package to version `{}`", newest.package_id);
                 Symlinker::new(self.config).set_active(self.register, &newest.package_id.clone(), installed_package.symlinked)?;
             }
         }
 
         // Delete the determined directory
+        if let Some(directory) = directory.to_str() {
+            debug!("Remove the package directory: {directory}");
```
```suggestion
            debug!("Removing the package directory: {directory}");
```

src/installer/installer.rs
```diff
@@ -433,28 +447,34 @@ impl<'a> Installer<'a> {
...
+        }
         self.remove_dir_all(&directory, &package_id.name)?;
 
-        // Remove package from installed package toml
-        self.register.remove_package_version(package_id);
+        // Remove package from the register
+        debug!("Remove {package_id} from the package register");
```
```suggestion
        debug!("Removing {package_id} from the package register");
```

src/installer/installer.rs
```diff
@@ -469,6 +489,7 @@ impl<'a> Installer<'a> {
         let directory = self.config.prefix_directory.join("packages").join(&package_name);
 
         // Remove active path symlink
+        debug!("Unlink the active path");
```
```suggestion
        debug!("Unlinking the active path");
```

src/installer/installer.rs
```diff
@@ -478,25 +499,32 @@ impl<'a> Installer<'a> {
         // Check if package was symlinked
         if let Some(package) = self.register.get_package(package_name) {
             if package.symlinked {
+                debug!("Unlink '{package_name}'");
```
```suggestion
                debug!("Unlinking '{package_name}'");
```

src/installer/installer.rs
```diff
@@ -478,25 +499,32 @@ impl<'a> Installer<'a> {
...
+        for package_version in &installed_versions {
             // Load source repository
             let repository = Repository::new(&package_version.source_repository_url, &package_version.source_repository_provider);
 
             // Run uninstall script
             self.run_uninstall_script(&repository, &package_version.package_id, &directory)?;
         }
 
+        if let Some(directory) = directory.to_str() {
+            debug!("Remove the package directory: {directory}");
```
```suggestion
            debug!("Removing the package directory: {directory}");
```

src/installer/installer.rs
```diff
@@ -478,25 +499,32 @@ impl<'a> Installer<'a> {
...
+        }
         self.remove_dir_all(&directory, package_name)?;
 
+        let uninstalled = installed_versions.iter().map(|p| p.package_id.clone()).collect();
+
         // Delete the installed package from toml
+        debug!("Remove {package_name} from the package register");
```
```suggestion
        debug!("Removing {package_name} from the package register");
```

src/installer/unpack.rs
```diff
@@ -47,12 +47,13 @@ impl ArchiveExtension {
 }
 
 // Unpacks files and saves them to the provided destination directory
-pub fn unpack<P: AsRef<Path>>(extension: ArchiveExtension, bytes: Bytes, destination_directory: P) -> Result<()> {
+pub fn unpack<P: AsRef<Path>>(package: String, extension: ArchiveExtension, bytes: Bytes, destination_directory: P) -> Result<()> {
```
BraveChicken1:
```suggestion
pub fn unpack<P: AsRef<Path>>(package: &PackageName, extension: ArchiveExtension, bytes: Bytes, destination_directory: P) -> Result<()> {
```

BraveChicken1 (changes requested):
Looks good overall!

src/cli/commands/install.rs
Max13245: num_args(1..) doesn't work for some reason, so I used required instead. I think it only works when the field is a flag. I'm not sure if this is the intended behaviour or a bug.

src/installer/installer.rs
Max13245: This might actually also be the case for the install script itself. A `PACKIT_VERBOSE` environment variable could work, this would mean we need to be careful with consistent logging throughout all scripts.

BraveChicken1: Yes, but we can enforce that through the core repository.

Max13245: Yes, I agree, I just mean that in that case it's more distributed and not managed at one place anymore. Which is what we need in this case, but it would mean we need to be a bit careful that it's the same across the entire core repository.

BraveChicken1 (approved):

PR #55, opened 2026-04-01
Add util checksum command

BraveChicken1:
This PR adds the `util checksum` command, calculating the checksum of a file at a given url.

### DIFF
--- a/src/cli/commands/util.rs
+++ b/src/cli/commands/util.rs
@@ -1,0 +1,68 @@
+use std::process::exit;
+
+use clap::Subcommand;
+use reqwest::Url;
+
+use crate::{
+    cli::{
+        commands::HandleCommand,
+        display::{Spinner, logging::error},
+    },
+    repositories::types::Checksum,
+};
+
+/// Provides several utils for advanced users.
+#[derive(Subcommand, Debug)]
+pub enum UtilArgs {
+    /// Calculates the checksum for the file at the given url
+    Checksum {
+        /// The url to request the file from
+        url: Url,
+    },
+}
+
+impl HandleCommand for UtilArgs {
+    /// Handles the util command.
+    fn handle(&self) {
+        match self {
+            Self::Checksum { url } => self.handle_checksum(url),
+        }
+    }
+}
+
+impl UtilArgs {
+    fn handle_checksum(&self, url: &Url) {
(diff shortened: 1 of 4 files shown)


src/cli/commands/util.rs
```diff
@@ -0,0 +1,68 @@
...
+        /// The url to request the file from
+        url: Url,
+    },
+}
+
+impl HandleCommand for UtilArgs {
+    /// Handles the util command.
+    fn handle(&self) {
+        match self {
+            Self::Checksum { url } => self.handle_checksum(url),
```
Max13245, line 29: Do we maybe also want to except PackageId as input? Would be really helpful if a certain url can't be reached/is not working at that time.

src/cli/commands/util.rs
```diff
@@ -0,0 +1,68 @@
...
+            Ok(response) => response,
+            Err(e) => {
+                spinner.stop();
+                error!(e, "Unable to request file");
+                exit(1);
+            },
+        };
+
+        if !response.status().is_success() {
+            spinner.stop();
```
Max13245: Maybe instead of doing a stop. Do a finish, but with the following text:
"Downloading file xyz unsuccessful"

Especially useful if we want to support PackageId input.

src/cli/commands/util.rs
```diff
@@ -0,0 +1,68 @@
...
+            Err(e) => {
+                spinner.stop();
+                error!(e, "Unable to get file bytes");
+                exit(1);
+            },
+        };
+
+        let checksum = Checksum::from_bytes(&bytes);
+
+        spinner.finish(format!("Found checksum {}", checksum.to_string()));
```
Max13245: I think the spinner should show content from the spinner, so: "Downloading xyz" and then "Downloading xyz successful". This clearly shows what's going on. This would also be more consistent with the rest of the repo.

Max13245 (commented):
Looks good overall. Do we maybe want to put util in a separate module? Or do we want to wait with that until there are more command utils?

src/cli/commands/util.rs
BraveChicken1: How would accepting PackageId as input work? Do you want to then fetch the url from the package metadata?

Max13245 (approved):

PR #67, opened 2026-04-18
Automatically set packit path in install.sh

Max13245:
This feature gives users the option to automatically add Packit to their PATH. 
It also fixes a bug which results when piping to the shell (which we do with curl). When piping there is no interactive shell, so now we check if an interactive shell is available, if not, we use /dev/tty with read for user input.

(With special thnx to Berke)

### DIFF
--- a/install.bat
+++ b/install.bat
@@ -20,6 +20,7 @@ mkdir "%PREFIX_DIR%\packages\packit\%VERSION%\bin"
 pushd "%PREFIX_DIR%\packages\packit\%VERSION%\bin"
 
 REM Install Packit to the prefix directory 
+echo Downloading Packit prebuild
 curl --proto "=https" -sSfL %SOURCE_PREBUILD_REPOSITORY_URL% --output packit.exe
 if not ERRORLEVEL 1 (
     echo Downloaded prebuild
@@ -163,6 +164,19 @@ if ERRORLEVEL 1 (
     echo Unsuccessfull install of Packit, the 'packit' command cannot be found
     popd
     exit /b 1
+)
+
+REM Ask the user if they want to automatically add Packit to their PATH
+set "answer="
+set /p "answer=Do you wish to automatically add Packit to your user PATH? (Y/n) "
+set "match="
+if "!answer!"=="y" set "match=1"
+if "!answer!"=="yes" set "match=1"
+if "!answer!"=="" set "match=1"
+if "!match!"=="1" (
+    setx PATH "%PATH%;%PREFIX_DIR%\bin"
+    popd
+    exit /b 0
 )
 
 echo Successfully installed Packit
--- a/install.bat
+++ b/install.bat
@@ -166,6 +166,14 @@ if ERRORLEVEL 1 (
     exit /b 1
 )
 
+echo Successfully installed Packit
+
+REM Exit early if Packit is already in the user PATH
+echo ";%PATH%;" | find /I ";%PREFIX_DIR%\bin;" >nul
+if %ERRORLEVEL%==0 (
+    exit /b 0
+)
+
 REM Ask the user if they want to automatically add Packit to their PATH
 set "answer="
 set /p "answer=Do you wish to automatically add Packit to your user PATH? (Y/n) "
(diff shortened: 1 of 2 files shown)


install.bat
```diff
@@ -165,6 +166,19 @@ if ERRORLEVEL 1 (
...
+if "!answer!"=="y" set "match=1"
+if "!answer!"=="yes" set "match=1"
+if "!answer!"=="" set "match=1"
+if "!match!"=="1" (
+    setx PATH "%PATH%;%PREFIX_DIR%\bin"
+    popd
+    exit /b 0
+)
+
 echo Successfully installed Packit
```
BraveChicken1: Shouldn't the succesfully installed message be printed before asking to add Packit to the PATH?

install.bat
```diff
@@ -165,6 +166,19 @@ if ERRORLEVEL 1 (
     exit /b 1
 )
 
+REM Ask the user if they want to automatically add Packit to their PATH
+set "answer="
+set /p "answer=Do you wish to automatically add Packit to your user PATH? (Y/n) "
```
BraveChicken1, line 179: Should we maybe also check if the path is already present in the PATH? From a previous installation for example.

install.sh
```diff
@@ -54,12 +55,20 @@ sudo chown -R $USERNAME "$PREFIX_DIR"
...
     echo "Retrieving prebuilds failed. Do you wish to build Packit from source? (Y/n)"
-    read answer
-    if [ $answer = "n" ] || [ $answer = "no" ]; then
+    
+    # Use /dev/tty when stdin is not available
+    if [ -t 0 ]; then
+        read answer
+    else
+        read answer < /dev/tty
+    fi
```
BraveChicken1: You could maybe use a function to avoid repeated code like this

install.sh
```diff
@@ -171,5 +193,55 @@ if ! command -v $PREFIX_DIR/bin/packit -h >/dev/null 2>&1; then
...
+
+    if [ "$answer" = "y" ] || [ "$answer" = "yes" ] || [ "$answer" = "" ]; then
+        echo "export PATH=\"$PREFIX_DIR/bin:\$PATH\"" >> "$SHELL_CONFIG_PATH"
+        exit 0
+    fi
+fi
+
+# If the shell is not recognized or user did not want to add Packit automatically tell the user how to add Packit to their shell config
+echo "Add $PREFIX_DIR/bin to your PATH by adding the command below to your shell config:"
```
BraveChicken1, line 229: Shouldn't this still contain the .bashrc and .zshrc notice?

BraveChicken1 (changes requested):
Looks good overall!

install.bat
Max13245: Yes, i'll do that.

install.sh
Max13245: yea, probably a good idea, I'll do that.

install.bat
Max13245: Yes, that's what we do now.

install.sh
Max13245: Well maybe,but I removed it because it's already in the message which asks if they want to install Packit automatically.

install.bat
BraveChicken1: No right? The succesfully installed message is printed on line 182, after asking if packit should be added to the PATH

Max13245: Oh yea, you're right, I was looking at install.sh instead of at install.bat.

install.bat
```diff
@@ -166,6 +167,20 @@ if ERRORLEVEL 1 (
...
+set "match="
+if "!answer!"=="y" set "match=1"
+if "!answer!"=="yes" set "match=1"
+if "!answer!"=="" set "match=1"
+if "!match!"=="1" (
+    setx PATH "%PATH%;%PREFIX_DIR%\bin"
+    popd
+    exit /b 0
+)
```
BraveChicken1, line 188: Could you also add the check if the path is already in the PATH variable for Windows?

BraveChicken1 (changes requested):

BraveChicken1 (approved):

PR #69, opened 2026-04-20
Add config command

BraveChicken1:
This PR adds a config command to change the config through the cli. It also moves the `repositories` command to `config repositories list`.


src/config.rs
```diff
@@ -118,3 +126,96 @@ impl Config {
...
+    pub fn add_to_repositories_rank(&mut self, repository_id: &str) {
+        let value = &mut self.document["repositories_rank"];
+        match value.is_none() {
+            true => {
+                let mut new_value = toml_edit::Array::new();
+                new_value.push(repository_id);
+                *value = new_value.into();
+            },
+            false => {
+                value.as_array_mut().expect("Expected repositories_rank to be an array!").push(repository_id);
```
Max13245, line 205: Since it's possible for users to edit their Config.toml manually we should probably throw a proper error instead of expecting it.

README.md
```diff
@@ -83,8 +80,27 @@ Packages the specified package into a prebuild and stores it in the destination
...
+Sets the multiuser setting to true or false. Currently not supported when there are already installed packages.
+
+#### `pit config repositories list`
+Lists all configured repositories.
+
+#### `pit config repositories set-rank <REPOSITORY-ID>`
+Sets the repositories rank in the config. Multiple `<REPOSITORY-ID>` can be given for multiple repositories in the rank.
+
+#### `pit config repositories add <ID> <URL> [PROVIDER]`
+Adds a new repository to the config. Also adds the new repository to the repositories rank.
```
Max13245:
```suggestion
Adds a new repository to the config. Also adds the new repository to the back of the repositories rank.
```

src/config.rs
Max13245, line 29: This is not necessarily for this PR, but it might be nice if we can combine the repositories and repositories_rank fields by using an ordered list of some sort.

src/cli/commands/config.rs
```diff
@@ -0,0 +1,211 @@
...
+                    continue;
+                },
+            };
+
+            // Print repository information
+            println!("{} ({repository_id})", metadata.name.bold().blue());
+            println!("{}", metadata.description.green());
+            println!("License: {}", metadata.license);
+            println!("Maintainers: {}", metadata.maintainers.join(", "));
+            println!("Repository provider: {}, path: {}", repository.provider, repository.path);
```
Max13245, line 179: When multiple repositories are printed it might be nice to have some kind of divider between them. Something simple like:
"--------------------------------------------------------------------"
will do. Or maybe just an empty line (that way we avoid having to adjust the line based on terminal width).

src/cli/commands/config.rs
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        config.set_repositories_rank(new_rank.clone());
+
+        config.save_to(&Config::get_default_path()).unwrap_or_exit_msg("Cannot save config file", 1);
+
+        println!("Succesfully set the repository rank to '{}'!", new_rank.join(", "));
+    }
+
+    /// Handles the config repositories add command
```
```suggestion
    /// Handles the config repositories add command.
```

src/cli/commands/config.rs
```diff
@@ -0,0 +1,211 @@
...
+            // Print repository information
+            println!("{} ({repository_id})", metadata.name.bold().blue());
+            println!("{}", metadata.description.green());
+            println!("License: {}", metadata.license);
+            println!("Maintainers: {}", metadata.maintainers.join(", "));
+            println!("Repository provider: {}, path: {}", repository.provider, repository.path);
+        }
+    }
+
+    /// Handles the config repositories set-rank command
```
```suggestion
    /// Handles the config repositories set-rank command.
```

src/cli/commands/config.rs
```diff
@@ -0,0 +1,211 @@
...
+
+        print!("Succesfully changed the multiuser setting to ");
+        if multiuser {
+            println!("on!");
+        } else {
+            println!("off!");
+        }
+    }
+
+    /// Handles the config repositories list command
```
```suggestion
    /// Handles the config repositories list command.
```

src/cli/commands/config.rs
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        config.set_prefix_directory(new_prefix.clone());
+
+        config.save_to(&Config::get_default_path()).unwrap_or_exit_msg("Cannot save config file", 1);
+
+        println!("Succesfully changed the prefix directory to {}!", new_prefix.display());
+    }
+
+    /// Handles the config set-multiuser command
```
```suggestion
    /// Handles the config set-multiuser command.
```

src/cli/commands/config.rs
```diff
@@ -0,0 +1,211 @@
...
+        if config.multiuser {
+            println!("on");
+        } else {
+            println!("off");
+        }
+
+        println!("Repositories rank: {}", config.repositories_rank.join(", "));
+    }
+
+    /// Handles the config set-prefix command
```
```suggestion
    /// Handles the config set-prefix command.
```

src/cli/commands/config.rs
```diff
@@ -0,0 +1,211 @@
...
+        /// The url of the new repository
+        url: String,
+
+        /// The optional provider of the new repository, `web` is used as default
+        provider: Option<String>,
+    },
+}
+
+impl HandleCommand for ConfigArgs {
+    /// Handles the config command
```
```suggestion
    /// Handles the config command.
```

src/cli/commands/config.rs
```diff
@@ -0,0 +1,211 @@
...
+    List,
+
+    /// Sets the repositories rank
+    SetRank {
+        /// The new rank to set
+        #[arg(required = true)]
+        new_rank: Vec<String>,
+    },
+
+    /// Adds a new repository to the config
```
Max13245, line 55: Please use periods after each sentence here as well, it matches the style of the other files.

src/cli/commands/config.rs
```diff
@@ -0,0 +1,211 @@
...
+            println!("{}", metadata.description.green());
+            println!("License: {}", metadata.license);
+            println!("Maintainers: {}", metadata.maintainers.join(", "));
+            println!("Repository provider: {}, path: {}", repository.provider, repository.path);
+        }
+    }
+
+    /// Handles the config repositories set-rank command
+    fn handle_set_repositories_rank(&self, mut config: EditableConfig, new_rank: &Vec<String>) {
+        for repo in new_rank {
```
Max13245, line 186: Do we also want to check for duplicate entries? This shouldn't give any problems if present and a user could still add them when editing Config.toml manually, but it might be neat. 

When using an ordered set we wouldn't need any of these checks.

Max13245 (changes requested):
Overall looks good! Just some small things.

Also, maybe change the get_default_path method to get_path in package_register.rs, because it's not a default path if we use the config to determine the prefix directory. 
I would say this isn't necessary for the get_default_path for the Config struct though, as this is independent from the prefix directory.

src/config.rs
BraveChicken1: The EditableConfig::from is the only way to create a EditableConfig. This function also uses the normal Config parsing to validate the config. I think this justifies the usage of expect here.

src/config.rs
BraveChicken1: I'm not sure if that will work, since the config can contain repositories that are not in the rank. This could be useful when we eventually support specifying the repository to install from in the install command.

src/cli/commands/config.rs
BraveChicken1: I'll add a check for duplicate entries, I think it's neat to deny this when using the commands to change the config.

src/cli/commands/config.rs
BraveChicken1: We already do this, I added a comment for clarity.

src/cli/commands/config.rs
BraveChicken1: For the command descriptions we don't use periods in the other files, so I'll leave them out for now.

src/cli/commands/config.rs
Max13245: Oops, missed it, sry.

src/cli/commands/config.rs
```diff
@@ -176,13 +177,21 @@ impl ConfigArgs {
         }
     }
 
-    /// Handles the config repositories set-rank command
+    /// Handles the config repositories set-rank command.
     fn handle_set_repositories_rank(&self, mut config: EditableConfig, new_rank: &Vec<String>) {
+        let mut seen = HashSet::new();
         for repo in new_rank {
+            // Check for invalid repository ids
```
Max13245, line 187: Please change both these println statements to actual errors.

Max13245 (changes requested):

Max13245 (approved):

PR #70, opened 2026-04-20
Add support for external test files

Max13245:
Feature which allows for external test file in the core repository.
Also fixes a bug in the read_file method in the filesystem implementation of MetadataProvider. We now convert the file_path parameter to a PathBuf to make it work on every platform.

### DIFF
--- a/src/installer/installer.rs
+++ b/src/installer/installer.rs
@@ -309,7 +309,27 @@ impl<'a> Installer<'a> {
                 &script_args,
                 self.options.verbose,
             );
-            scripts::run_test_script(&script_data)?;
+
+            let external_files = version_meta.external_test_files.iter().chain(&target.external_test_files);
+            let mut read_files = Vec::new();
+            let mut skip_test = false;
+            for file in external_files {
+                let file_content =
+                    self.repository_manager.read_file(&install_meta.repository_id, &install_meta.package_metadata.name, file)?;
+
+                match file_content {
+                    Some(content) => read_files.push((file, content)),
+                    None => {
+                        warning!("Skipping test, because the required files could not be downloaded.");
+                        skip_test = true;
+                        break;
+                    },
+                }
+            }
+
+            if !skip_test {
+                scripts::run_test_script(&script_data, &read_files)?;
+            }
         }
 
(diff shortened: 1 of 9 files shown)


src/installer/installer.rs
```diff
@@ -309,7 +309,27 @@ impl<'a> Installer<'a> {
...
+                        warning!("Skipping test, because the required files could not be downloaded.");
+                        skip_test = true;
+                        break;
+                    },
+                }
+            }
+
+            if !skip_test {
+                scripts::run_test_script(&script_data, &read_files)?;
+            }
```
BraveChicken1, line 332: This functionality could maybe be moved to the run_test_script function, then you could also use an early return instead of the skip_test variable.

src/repositories/types/package_target.rs
```diff
@@ -26,4 +26,13 @@ pub struct PackageTarget {
     pub postinstall_script: Option<Script>,
     pub test_script: Option<Script>,
     pub uninstall_script: Option<Script>,
+
+    #[serde(default = "PackageTarget::default_external_files")]
```
BraveChicken1:
```suggestion
    #[serde(default)]
```

src/repositories/types/package_version.rs
```diff
@@ -51,6 +51,9 @@ pub struct PackageVersionMeta {
     #[serde(default = "PackageVersionMeta::default_use_version_specific")]
     pub use_version_specific_uninstall: bool,
 
+    #[serde(default = "PackageVersionMeta::default_external_files")]
```
BraveChicken1:
```suggestion
    #[serde(default)]
```

src/repositories/manager.rs
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: GPL-3.0-only
-use std::collections::HashMap;
+use std::{collections::HashMap, path::PathBuf};
```
BraveChicken1: I think we don't need this import here

BraveChicken1 (changes requested):

BraveChicken1 (approved):

PR #71, opened 2026-04-20
Add fuzzy search

Max13245:


src/cli/commands/check.rs
```diff
@@ -31,7 +31,24 @@ impl HandleCommand for CheckArgs {
...
+                    let versions = package_name.versions.keys();
+                    print!("Did you mean version(s): ");
+                    for version in versions {
+                        print!("'{version}' ");
+                    }
+                    println!();
+                    return;
+                }
+
+                let fuzzy_match = min_fuzzy_search(register.iterate_package_names(), &package_id.name);
```
BraveChicken1:
```suggestion
                let fuzzy_match = fuzzy::min_fuzzy_search(register.iterate_package_names(), &package_id.name);
```

I would specify the module when using a function from a different module directly. Then it's immediately clear the function is specified in a different file. You could then even rename the `min_fuzzy_search` function.

src/cli/commands/link.rs
```diff
@@ -1,14 +1,19 @@
+use std::process::exit;
+
 // SPDX-License-Identifier: GPL-3.0-only
```
BraveChicken1: The SPDX identifier is moved down, please move it back up

src/cli/commands/switch.rs
```diff
@@ -44,10 +59,25 @@ impl HandleCommand for SwitchArgs {
...
 
         // Get installed package version
-        let package_version = package.get_package_version(&self.package_version).unwrap_or_exit_msg(
-            &format!("Package {}@{} is not installed!", self.package_name, self.package_version),
-            1,
-        );
+        let package_version = match package.get_package_version(&self.package_version) {
+            Some(package_version) => package_version,
+            None => {
+                error!(msg: "Package '{}' cannot be found.", self.package_version);
```
BraveChicken1:
```suggestion
                error!(msg: "Package '{}@{}' cannot be found.", self.package_name, self.package_version);
```

src/utils/fuzzy.rs
```diff
@@ -0,0 +1,177 @@
...
+    Ok(best_word)
+}
+
+/// Wraps around the `fuzzy_search` method and gets the fuzzy match with
+/// the lowest distance to the given string.
+/// Returns None if there are no fuzzy matches and a `PackageName` if there is at least one fuzzy match.
+pub fn min_fuzzy_search<'a, I>(words: I, string: &str) -> Option<PackageName>
+where
+    I: IntoIterator,
+    I::Item: AsRef<PackageName>,
```
BraveChicken1: Couldn't this just be `I::Item: &PackageName`? Then you wouldn't need the AsRef implementation for PackageName.

BraveChicken1 (changes requested):

src/utils/fuzzy.rs
Max13245: This would be possible for the min_fuzzy_search, but not for the fuzzy_search, because in the repository_fuzzy_search (and possibly at other method in the future) we pass a HashSet<PackageName>, which has items which are not references (and difficult to efficiently convert/pass by reference).

BraveChicken1: You can use `.iter()` to convert the HashSet into an iterator containing references

src/cli/commands/search.rs
```diff
@@ -27,7 +27,13 @@ impl HandleCommand for SearchArgs {
         let (repository_id, package) = match manager.read_package(&self.optional_id.name) {
             Ok(package) => package,
             Err(RepositoryError::PackageNotFoundError { .. }) => {
-                println!("Cannot find package {}", self.optional_id.name);
+                println!("Cannot find package '{}'", self.optional_id.name);
+
+                let fuzzy_match = repository_fuzzy_search(&config, &manager, &self.optional_id.name).unwrap_or_exit(1);
```
BraveChicken1: Could you also reference the `repository_fuzzy_search` with the `fuzzy::` prefix?

src/cli/commands/unlink.rs
```diff
@@ -1,12 +1,14 @@
+use std::process::exit;
+
 // SPDX-License-Identifier: GPL-3.0-only
```
BraveChicken1: This file also has the SPDX identifier problem

BraveChicken1 (changes requested):

BraveChicken1:
Please also update the CHANGELOG to contain the new feature.

BraveChicken1 (approved):

PR #72, opened 2026-04-21
Add leading zeros support for versions

Max13245:


src/installer/types/version.rs
```diff
@@ -158,14 +149,18 @@ impl FromStr for Version {
...
+        }
     }
 }
 
 /// Implements the from trait for `&[u32; N]`.
 impl<const N: usize> From<&[u32; N]> for Version {
     fn from(value: &[u32; N]) -> Self {
-        Self { numbers: Vec::from(value) }
+        Self {
+            numbers: value.iter().map(|v| VersionNumber::from(v.clone())).collect(),
```
BraveChicken1, line 171: I think you could even use `value.iter().cloned().map(Into::into).collect()` or even `value.iter().cloned().collect()` here.

src/installer/types/version.rs
```diff
@@ -10,18 +10,17 @@ use std::{
 use serde::{Deserialize, Serialize, de};
 use thiserror::Error;
 
+use crate::installer::types::version_number::VersionNumber;
```
BraveChicken1, line 13: I would use a pub use in the mod.rs, then this import can be just `crate::installer::types::VersionNumber`

src/installer/types/version.rs
```diff
@@ -139,16 +139,7 @@ impl FromStr for Version {
 
         let mut version_parts = Vec::new();
         for num in string.split('.') {
-            if num.is_empty() {
-                return Err(VersionError::DotsError);
-            }
```
BraveChicken1, line 144: I think having the empty check here too is neater, because then we can differentiate between an empty version or just multiple consecutive dots.

src/installer/types/version_number.rs
```diff
@@ -0,0 +1,147 @@
...
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        if self.number > other.number {
+            return Ordering::Greater;
+        }
+
+        if self.number < other.number {
+            return Ordering::Less;
+        }
+
+        Ordering::Equal
```
BraveChicken1, line 23: Is it correct that we ignore leading zeros for comparing? I'm not fully sure if this will not cause any bugs in other parts of the application.

BraveChicken1 (changes requested):

src/installer/types/version.rs
Max13245: The VersionNumber is never used outside of the types module though. I think it's nice to keep it only available for the components that need it.

src/installer/types/version_number.rs
Max13245: Well that's the whole point isn't it? Leading zeros are mostly a stylistic choice (which we want to support), but we shouldn't evaluate them, as they have no meaning.

src/installer/types/version.rs
Max13245: I'll take a look.

src/installer/types/version.rs
Max13245: Mhm, yea I wasn't sure about this, because consecutive dots is the same as a version number which is none. If I implement the DotsError check, then the NoneError will never be used by the version number. So I thought it would be neater to have the check at one place.

src/installer/types/version.rs
Max13245: `value.iter().cloned().collect()` doesn't work. `value.iter().cloned().map(Into::into).collect()` does, but is less explicit (and not that much shorter), so I would rather leave this as is.

src/installer/types/version.rs
BraveChicken1: The empty version number now uses the same error message as an empty version. You could either add a empty version number error, or a consecutive dots error. For the last one you could simply map the NoneError from the VersionNumber to DotsError.

Max13245: True, I'll do that.

BraveChicken1 (approved):

PR #73, opened 2026-04-22
Add web prebuild provider

Max13245:


BraveChicken1 (approved):
I tested it and fixed some issues. I also changed the expected path a bit:
`packages/<target>/<prefix>/<package-name>/<file-name>`
This file path should maybe also become the default of the FileSystemPrebuildProvider.

src/repositories/error.rs
```diff
@@ -38,6 +40,9 @@ pub enum RepositoryError {
     #[error("No supported version for the current target could be found for package '{0}'.")]
     SupportError(String),
 
+    #[error("Request returned an unsuccessful status code '{0}'.")]
+    UnsuccessfulRequest(StatusCode),
```
BraveChicken1:
```suggestion
    UnsuccessfulRequest(reqwest::StatusCode),
```

src/repositories/error.rs
```diff
@@ -49,6 +54,9 @@ pub enum RepositoryError {
 
     #[error("Cannot parse checksum from hex")]
     ChecksumParseError(#[from] hex::FromHexError),
+
+    #[error("Cannot parse url")]
+    UrlParseError(#[from] ParseError),
```
BraveChicken1:
```suggestion
    UrlParseError(#[from] url::ParseError),
```

BraveChicken1 (changes requested):

BraveChicken1 (approved):

PR #74, opened 2026-04-24
Add support for patches to packages

BraveChicken1:
This PR adds support for applying patches to the source code of packages.

Please note that this cannot be merged yet, because diffy did not yet release the version containing multifile patch support.

### DIFF
--- a/src/installer/builder.rs
+++ b/src/installer/builder.rs
@@ -161,9 +161,11 @@ impl<'a> Builder<'a> {
             apply_directory = apply_directory.join(PathBuf::from(apply_in));
         }
 
+        let package_id = PackageId::new(package_name.clone(), version.clone());
+
         // Apply patches
         for (id, patch) in source.get_sorted_patches() {
-            let description = format!("patch {id}' of '{package_name}");
+            let description = format!("patch {id}' of '{package_id}");
             let patch_bytes = self.download_file(&patch.url, &patch.mirrors, &patch.checksum, &description)?;
 
             // Construct apply directory for this patch
@@ -175,7 +177,7 @@ impl<'a> Builder<'a> {
             // Apply patch
             patches::apply_patch(&patch_bytes, &apply_directory)?;
 
-            println!("Applied patch '{id}' to '{package_name}'");
+            println!("Applied patch '{id}' to '{package_id}'");
         }
 
         // Create build env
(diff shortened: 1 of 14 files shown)


src/installer/builder.rs
Max13245: Maybe move the package id (one of them, delete the other) to top. Then it can be used for the patch applied print after a patch:
```
println!("Applied patch '{id}' to '{package_id}'");
```

src/repositories/types/common.rs
```diff
@@ -29,12 +29,68 @@ pub struct Source {
...
+    pub apply_patches_in: Option<String>,
+
+    #[serde(default, deserialize_with = "Source::deserialize_patches")]
+    pub patches: HashMap<u32, Patch>,
 }
 
 /// Wrapper to differentiate between Single and Named sources in the metadata files.
-#[derive(Deserialize, Debug)]
-#[serde(untagged)]
+#[derive(Debug)]
```
Max13245, line 39: Don't we need the serde(untagged), because of the different variant types?

Edit:
Now I see we have a custom implementation, is there a specific reason for this?

src/repositories/types/common.rs
```diff
@@ -29,12 +29,68 @@ pub struct Source {
...
+
+    #[serde(default)]
+    pub mirrors: Vec<String>,
+    pub apply_in: Option<String>,
+}
+
+impl Source {
+    /// Gets all patches of the source, sorted by id.
+    pub fn get_sorted_patches(&self) -> &HashMap<u32, Patch> {
+        &self.patches //TODO
```
Max13245: This is still TODO (This should probably also be done before a merge), but it returns a HashMap for a sorted set. Which is not possible.

src/utils/patches.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    let original = match source {
+        Some(source) => fs::read(&source)?,
+        None => vec![],
+    };
+
+    let patched = match patch {
+        PatchKind::Text(patch) => diffy::apply_bytes(&original, &patch)?,
+        PatchKind::Binary(patch) => patch.apply(&original)?,
+    };
```
Max13245, line 124: What is the behaviour when vec![] is passed? Will the patch just be skipped?

Max13245:
Looks good! Just some small check/changes.

Also did we test this on Windows already?

src/repositories/types/common.rs
BraveChicken1: Previously the Source struct did not contain any subtables, so serde was able to differentiate between a Single source (just a table) and a Named source (a table with subtables). Because we now have patches as a subtable, we need to have a custom deserialize implementation to ensure serde correctly recognizes the single and named variants. When doing a custom implementation, the untagged attribute is not applied anymore, so we can remove it.

src/utils/patches.rs
BraveChicken1: When vec![] is passed, this means there is no original. This is used when the patch creates a new file. The example of diffy also uses this approach.

src/installer/builder.rs
BraveChicken1: I will change it. The duplicate package id is probably a leftover from a merge somewhere, I'll fix that too.

src/repositories/types/common.rs
BraveChicken1: Yes, I forgot to do this, I will fix this.

BraveChicken1:
It's not yet tested on Windows, that should probably be done before merging. I will commit my last changes and after that we can test it on Windows too.

Max13245:
Has this been tested on Windows now? If so, I'll approve it.

BraveChicken1:
It fully works on Windows!

Max13245:
Great!

Max13245 (approved):

PR #76, opened 2026-04-25
Add --updatables flag and grid print to list command

Max13245:


src/cli/display/grid.rs
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: GPL-3.0-only
+use terminal_size::{Height, Width, terminal_size};
+
+/// Prints the given items in a grid.
+pub fn print_grid(items: Vec<String>) {
+    // Get the width of the terminal for grid calculations
+    let terminal_width = match terminal_size() {
+        Some((Width(width), Height(_))) => width as usize,
+        None => panic!("TODO"),
```
BraveChicken1: If you don't have a terminal width, I would just default to printing it in a long list (how it was before)

src/cli/display/grid.rs
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: GPL-3.0-only
+use terminal_size::{Height, Width, terminal_size};
+
+/// Prints the given items in a grid.
+pub fn print_grid(items: Vec<String>) {
```
BraveChicken1:
```suggestion
pub fn print_grid<T>(items: Vec<T>)
where
  T: Display
{
```

If you accept any list of displayable items here, you don't need the to_string in the list command

BraveChicken1 (changes requested):
Looks good overall!

src/cli/display/grid.rs
Max13245: Yea good idea. I forgot that TODO. I'll fix that.

src/cli/display/grid.rs
Max13245: The problem with this is that I need the `len` method. So then I would need to add another trait to `T`.  `AsRef<str>` or `ExactSizeIterator`, for both of these it doesn't really help with avoiding the `to_string` method. 

Alternatively I could create a wrapper method which has `Vec<PackageId>` as parameter.

BraveChicken1: When a type implements the Display trait, you should have the `to_string` available. You could then use this inside the print_grid to transform it to strings.

Max13245: Yes, that should work.

BraveChicken1 (approved):

PR #78, opened 2026-05-02
Add build-install xtask

BraveChicken1:
This PR adds the build-install xtask to Packit, which can be used to build and install Packit.

The xtask runs the `cargo build --release` command and copies it into `<destination>/bin`. It also creates the `pit` symlink in the bin directory.

Please note that the install scripts need to be adjusted to use this new xtask before releasing v0.0.2.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 /target
-Config.toml
 /temp
 .DS_Store
 .vscode/
--- a/xtasks/xtask-build-install/src/main.rs
+++ b/xtasks/xtask-build-install/src/main.rs
@@ -1,0 +1,115 @@
+use std::{
+    fs,
+    path::{Path, PathBuf},
+    process::{Command, exit},
+};
+
+use clap::Parser;
+
+#[derive(Parser, Debug)]
+#[command(name = "build-install", version, about)]
+pub struct TaskCommand {
+    /// The destination to install to
+    #[arg(long)]
+    destination: Option<PathBuf>,
+}
+
+fn main() {
+    let command = TaskCommand::parse();
+
+    // Build with cargo
+    run_build();
+
+    // Retrieve destination
+    let destination = command.destination.unwrap_or(get_target_path().join("build"));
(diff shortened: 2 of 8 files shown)


xtasks/xtask-build-install/src/main.rs
```diff
@@ -0,0 +1,115 @@
...
+    // Create symlink for pit to packit
+    if let Err(e) = create_file_symlink("packit", bin_directory.join("pit")) {
+        eprintln!("Failed to create pit symlink in bin directory");
+        eprintln!("Error: {e}");
+        exit(1);
+    }
+}
+
+fn get_project_root_path() -> PathBuf {
+    Path::new(std::env!("CARGO_MANIFEST_DIR")).ancestors().nth(2).unwrap().to_path_buf()
```
Max13245: Please handle this unwrap method properly.

xtasks/xtask-build-install/src/main.rs
```diff
@@ -0,0 +1,115 @@
...
+        },
+        Err(e) => {
+            eprintln!("Failed to build Packit, failed command: `{cmd:?}`");
+            eprintln!("Error: {e}");
+            exit(1);
+        },
+    }
+}
+
+fn copy_binary_to_destination(destination: &PathBuf) {
```
Max13245, line 72: Please add rust doc to this method (and the other methods in this file).

.gitignore
```diff
@@ -1,5 +1,4 @@
 /target
-Config.toml
 /temp
```
Max13245, line 2: If we're deleting stuff from the .gitignore, maybe also remove the /temp directory.

Max13245 (changes requested):
Looks good, just a couple small changes.

.gitignore
BraveChicken1: I kept the /temp directory, since it could maybe still be useful for testing in the future. An actual module in the repository should also never be called "temp" in my opinion.

Max13245 (approved):

PR #79, opened 2026-05-02
Add portable repositories

BraveChicken1:
This PR adds support for generating portable repositories which can be used to install packages on airgapped systems.

Prebuilds are also downloaded into the portable repo, but is not automatically configured as prebuild repository in the repository.toml file. This is because there is currently no way to have prebuild repositories inside metadata repositories, in the future we should probably implement a new `relative` prebuild provider, which uses the metadata provider to download the prebuilds.


src/repositories/portable_repo.rs
Max13245, line 1: There should be functionality which adds dependencies of a package and in case of no prebuilds also the build dependencies. Otherwise an added packages would never work and there would be no purpose for adding it. 

If you want to keep this for flexibility reasons I would suggest a `--disable-dependencies` flag. However, I don't think this should be the default behaviour.

Max13245 (changes requested):
Looks really good!

src/repositories/portable_repo.rs
Max13245: You can probably just use the install tree functionality to get the dependencies.

BraveChicken1: That would indeed be a good addition. Build dependencies would only be included on the --exclude-prebuilds flag. I'll take a look at implementing this.

src/cli/commands/util.rs
```diff
@@ -37,6 +37,10 @@ pub enum UtilArgs {
         /// True to exclude prebuilds in the portable repository
         #[arg(long, default_value = "false")]
         exclude_prebuilds: bool,
+
+        /// True to skip automatic addition fo dependencies to the portable repository
```
Max13245: /// True to skip automatic addition for dependencies to the portable repository

src/cli/commands/util.rs
```diff
@@ -91,15 +96,29 @@ impl UtilArgs {
         println!("Found checksum {}", checksum.to_string());
     }
 
-    fn handle_portable_repo(&self, destination: &PathBuf, packages: &Vec<PackageId>, exclude_prebuilds: bool) {
+    fn handle_portable_repo(
+        &self,
+        destination: &PathBuf,
+        packages: &Vec<PackageId>,
+        exclude_prebuilds: bool,
```
Max13245, line 103: Currently the list of flags is quite short, but I could see it being larger in the future. It might be nice to have a PortableRepoOption struct to pass flags (like we have in the installer). This is currently not yet necessary though.

src/repositories/portable_repo.rs
```diff
@@ -164,6 +159,61 @@ impl<'a> PortableRepoCreator<'a> {
         Ok(())
     }
 
+    /// Creates a set of all packages that are in the dependency trees of the included packages.
+    /// Checks for multiple repository conflicts.
```
Max13245: It might be nice to give a bit more explanation here.
/// Checks for multiple repository conflicts. Where a conflict is two (or more) versions of the same package, but both from a different repository.
/// Returns a `PortableRepoError::PackageFromMultipleRepositories` in case of a repository conflict.

src/repositories/portable_repo.rs
```diff
@@ -164,6 +159,61 @@ impl<'a> PortableRepoCreator<'a> {
...
+            }
+
+            let package_name = package_id.name.clone();
+
+            let (repository_id, package_version) = self.repository_manager.read_package_version(&package_id, &self.target)?;
+
+            // Check if the package name was previously requested from a different repository
+            if package_index.contains_key(&package_name) && package_index[&package_name] != repository_id {
+                return Err(PortableRepoError::PackageFromMultipleRepositories { package_name });
+            }
```
Max13245, line 187: This is probably a bit overkil, so feel free not to implement this. However, it might be nice to check if the previously found package can also be found in the currently found repository. Then we don't have to throw the error. Note that we wouldn't have to check if the current package can be found in the previously found repository, because of the repository rank.

src/repositories/portable_repo.rs
```diff
@@ -164,6 +159,61 @@ impl<'a> PortableRepoCreator<'a> {
...
+
+            // Add package data to package set and index
+            all_packages.insert(package_id);
+            package_index.insert(package_name, repository_id);
+
+            // Skip dependency resolution if the flag is enabled
+            if self.skip_dependency_resolution {
+                continue;
+            }
```
Max13245, line 196: I would move this check out of this function and only use this method if skip_dependency_resolution is false. Then it doesn't have to be checked every iteration.

Or did you do this to also do the repository check for the given packages?

Max13245 (changes requested):

src/cli/commands/util.rs
BraveChicken1: In the future I want to refactor the util command into separate files, then we could just use a `PortableRepoArgs` struct here for all command arguments.

src/repositories/portable_repo.rs
BraveChicken1: Hmmm, I don't necessarily dislike this idea, but I'm not fully sure if this is wanted behaviour, since the repositories_rank is not used anymore then.

src/repositories/portable_repo.rs
BraveChicken1: Yes, we always want to do the repository check, so I chose to do it all in the same loop. I could split out this loop, but I'm not sure if we want that?

src/repositories/portable_repo.rs
Max13245: That's true, or at least half true, of course a repository with a higher ranking (higher being the first repo) will still be preferred. For example in case of 3 repo's where the first only matches 1 of two packages and the last two match both. In this case the higher ranked repo will still be used. 

However, I do get your point as well and I think this is not necessarily for this PR.

src/repositories/portable_repo.rs
Max13245: I guess it's fine. It's only slightly more inefficient, because of the creation of a new Vec.

Max13245 (approved):

PR #81, opened 2026-05-05
Verifier structural changes and new checks

Max13245:
- The check and fix command now except multiple package parameters.
- Add support for 'check dependencies'.
- The inconsistent register fix doesn't require a re-install anymore.
- Add check after fix (re-run check to make sure the fix worked).
- Add checks and fixes for: stray directories, empty directories, package existence, correct permissions, missing Config.toml and missing Installed.toml.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - The list command now has grid printing.
 - The regex search, to search packages based on a given regex pattern.
 - Add support for applying patches to the source code of packages.
+- Verifier and Repairer features
+    - The check and fix command now except multiple package parameters.
+    - Add support for 'check dependencies' (checks which have to be executed before the current check).
+    - The inconsistent register fix doesn't require a re-install anymore.
+    - Add check after fix (re-run check to make sure the fix worked).
+    - Add checks and fixes for: stray directories, empty directories, package existence, correct permissions, missing Config.toml and missing Installed.toml.
 
 ### Changes
 - The build system now includes a new `build-install` xtask to create a full Packit build in a `bin` directory structure.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - The regex search, to search packages based on a given regex pattern.
 - Add support for applying patches to the source code of packages.
 - Verifier and Repairer features
-    - The check and fix command now except multiple package parameters.
-    - Add support for 'check dependencies' (checks which have to be executed before the current check).
+    - The check and fix command now accept multiple package parameters.
     - The inconsistent register fix doesn't require a re-install anymore.
     - Add check after fix (re-run check to make sure the fix worked).
     - Add checks and fixes for: stray directories, empty directories, package existence, correct permissions, missing Config.toml and missing Installed.toml.
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Shows info about the specified installed package. If the `-v` option is given, e
 #### `pit check [<PACKAGE-NAME>@<VERSION> ...]`
 Checks the Packit installation for issues. When package name(s) and version(s) are given, only those package(s) are checked for issues. 
 
-#### `pit fix`
+#### `pit fix [<PACKAGE-NAME>@<VERSION> ...]`
 Fix all issues found by the check command. You will be asked if you want to fix an issue for each issue type. When package name(s) and version(s) are given, only those package(s) are checked and fixed. 
 
 #### `pit switch <PACKAGE-NAME> <VERSION> [--skip-symlinking]`
--- a/README.md
+++ b/README.md
@@ -61,11 +61,11 @@ Updates the specified package to the new version, or the latest version if no ne
 #### `pit info <PACKAGE-NAME>[@<VERSION>] [-v] [--tree]`
 Shows info about the specified installed package. If the `-v` option is given, extra information is shown. If the `--tree` option is enabled, the whole dependency tree is shown.
 
-#### `pit check [<PACKAGE-NAME>@<VERSION>]`
-Checks the Packit installation for issues. When a package name and version is given, only that package is checked for issues.
+#### `pit check [<PACKAGE-NAME>@<VERSION> ...]`
+Checks the Packit installation for issues. When package name(s) and version(s) are given, only those package(s) are checked for issues. 
 
 #### `pit fix`
-Fix all issues found by the check command. You will be asked if you want to fix an issue for each issue type.
+Fix all issues found by the check command. You will be asked if you want to fix an issue for each issue type. When package name(s) and version(s) are given, only those package(s) are checked and fixed. 
 
 #### `pit switch <PACKAGE-NAME> <VERSION> [--skip-symlinking]`
 Switches the active version of the specified package to the specified version. If the `--skip-symlinking` option is given, the new active version is not symlinked into the /bin, /lib, /share, etc. directories.
--- a/src/cli/commands/check.rs
+++ b/src/cli/commands/check.rs
@@ -15,8 +15,8 @@ use crate::{
 /// Checks for any inconsistencies or mistakes in the installed packages or in the Packit files itself.
 #[derive(Args, Debug)]
 pub struct CheckArgs {
-    /// Optional package id, to limit the check to the specified package
-    package: Option<PackageId>,
+    /// A vec of packages to fix. Could be empty, then all packages are checked.
+    pub packages: Vec<PackageId>,
 }
 
 impl HandleCommand for CheckArgs {
@@ -26,20 +26,37 @@ impl HandleCommand for CheckArgs {
         let register = PackageRegister::from(&register_dir).unwrap_or_exit(1);
         let mut verifier = Verifier::new(&config);
 
-        // Check if the package exists before checking it with the verifier
-        if let Some(package_id) = &self.package {
+        let error_message = "An error occured during the check, this error could be caused by one of the issues above and might still be fixed by `pit fix`. It's possible that not all issues were found (especially when checking a single package).";
+        if self.packages.is_empty() {
+            while let Some(issue) = verifier.next_issue(&register).unwrap_or_exit_msg(error_message, 1) {
+                println!("{issue}")
+            }
+
+            // Return correct message based on found issues
+            if verifier.issues_found() {
+                println!("Consider running `pit fix` to resolve the issues above.");
+            } else {
+                println!("No issues were found");
+            }
+
+            return;
+        }
+
+        for package_id in &self.packages {
             // Check if the package exists in the register or in storage before doing any checks
+            // TODO: The verifier should handle this (at the places where it currently says "note doesn't check for existance in xyz")
             let installed_directory = config.prefix_directory.join("packages").join(&package_id.name).join(package_id.version.to_string());
             if register.get_package_version(package_id).is_none() && !fs::exists(installed_directory).unwrap_or_exit(1) {
                 error!(msg: "Cannot perform checks, package '{package_id}' doesn't exist in register or storage.");
 
                 // Show possible versions if a package with the given name exists
                 if let Some(package_name) = register.get_package(&package_id.name) {
-                    let versions = package_name.versions.keys();
                     print!("Did you mean version(s): ");
+                    let versions = package_name.versions.keys();
                     for version in versions {
                         print!("'{version}' ");
                     }
+
                     println!();
                     return;
                 }
@@ -51,28 +68,17 @@ impl HandleCommand for CheckArgs {
 
                 exit(1);
             }
-        }
 
-        // Show all issues
-        let error_message = "An error occured during the check, this error could be caused by one of the issues above and might still be fixed by `pit fix`. It's possible that not all issues were found.";
-        match &self.package {
-            Some(id) => {
-                while let Some(issue) = verifier.next_package_issue(id, &register).unwrap_or_exit_msg(error_message, 1) {
-                    println!("{issue}")
-                }
-            },
-            None => {
-                while let Some(issue) = verifier.next_issue(&register).unwrap_or_exit_msg(error_message, 1) {
-                    println!("{issue}")
-                }
-            },
-        }
+            while let Some(issue) = verifier.next_package_issue(package_id, &register).unwrap_or_exit_msg(error_message, 1) {
+                println!("{issue}")
+            }
 
-        // Return correct message based on found issues
-        if verifier.issues_found() {
-            println!("Consider running `pit fix` to resolve the issues above.");
-        } else {
-            println!("No issues were found");
+            // Return correct message based on found issues
+            if verifier.issues_found() {
+                println!("Consider running `pit fix` to resolve the issues above.");
+            } else {
+                println!("No issues were found");
+            }
         }
     }
 }
--- a/src/cli/commands/check.rs
+++ b/src/cli/commands/check.rs
@@ -37,8 +37,6 @@ impl CheckArgs {
         if verifier.issues_found() {
             println!("{ISSUE_FOUND_MESSAGE}");
             return;
-        } else {
-            println!("{NO_ISSUE_FOUND_MESSAGE}");
         }
 
         let config = Config::from(&Config::get_default_path()).unwrap_or_exit_msg("Cannot load config", 1);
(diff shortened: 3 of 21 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - The list command now has grid printing.
 - The regex search, to search packages based on a given regex pattern.
 - Add support for applying patches to the source code of packages.
+- Verifier and Repairer features
+    - The check and fix command now except multiple package parameters.
```
```suggestion
    - The check and fix command now accept multiple package parameters.
```

CHANGELOG.md
```diff
@@ -16,6 +16,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
+    - Add support for 'check dependencies' (checks which have to be executed before the current check).
```
BraveChicken1: I'm not sure if we should include this one in the changelog, since it's mainly internal behaviour and not something that the user will really see I think

README.md
```diff
@@ -61,11 +61,11 @@ Updates the specified package to the new version, or the latest version if no ne
 #### `pit info <PACKAGE-NAME>[@<VERSION>] [-v] [--tree]`
 Shows info about the specified installed package. If the `-v` option is given, extra information is shown. If the `--tree` option is enabled, the whole dependency tree is shown.
 
-#### `pit check [<PACKAGE-NAME>@<VERSION>]`
-Checks the Packit installation for issues. When a package name and version is given, only that package is checked for issues.
+#### `pit check [<PACKAGE-NAME>@<VERSION> ...]`
```
BraveChicken1, line 64: We currently don't use this `...` syntax for other commands. This is maybe also a nice addition for the other commands that accept multiple package ids

README.md
```diff
@@ -61,11 +61,11 @@ Updates the specified package to the new version, or the latest version if no ne
...
+Checks the Packit installation for issues. When package name(s) and version(s) are given, only those package(s) are checked for issues. 
 
 #### `pit fix`
```
```suggestion
#### `pit fix [<PACKAGE-NAME>@<VERSION> ...]`
```

src/cli/commands/check.rs
```diff
@@ -1,78 +1,78 @@
...
+    cli::commands::HandleCommand, config::Config, installer::types::PackageId, storage::package_register::PackageRegister,
+    utils::unwrap_or_exit::UnwrapOrExit, verifier::Verifier,
 };
 
 /// Checks for any inconsistencies or mistakes in the installed packages or in the Packit files itself.
 #[derive(Args, Debug)]
 pub struct CheckArgs {
-    /// Optional package id, to limit the check to the specified package
-    package: Option<PackageId>,
+    /// A vec of packages to check. Could be empty, then all packages are checked.
```
BraveChicken1:
```suggestion
    /// A list of packages to check. Could be empty, then all packages are checked.
```

src/cli/commands/fix.rs
```diff
@@ -15,32 +18,76 @@ use crate::{
 
 /// Fixes all issues found by the check command. The user will be asked if they want to fix an issue for each issue type.
 #[derive(Args, Debug)]
-pub struct FixArgs;
+pub struct FixArgs {
+    /// A vec of packages to fix. Could be empty, then all packages are fixed.
```
BraveChicken1:
```suggestion
    /// A list of packages to fix. Could be empty, then all packages are fixed.
```

src/cli/commands/fix.rs
```diff
@@ -15,32 +18,76 @@ use crate::{
...
+
+impl FixArgs {
+    /// Fixes all packages.
+    fn fix_all(&self, verifier: &mut Verifier, repairer: &mut Repairer) {
+        let mut issue_index = -1;
+        while let Some(issue) = verifier.next_initial_issue().unwrap_or_exit(1) {
+            if verifier.get_initial_check_index() as i32 - 1 == issue_index {
+                error!(msg: UNSUCCESSFUL_FIX_MESSAGE);
+                exit(1);
+            }
```
BraveChicken1, line 50: What does this do? Could you maybe add a comment here?

src/platforms/permissions/unix.rs
```diff
@@ -70,6 +70,11 @@ pub fn set_packit_permissions(path: &PathBuf, is_multiuser: bool, recurse: bool)
     set_file_permissions(path, metadata.permissions(), group_id, recurse)
 }
 
+/// Checks if the packit group exists.
+pub fn packit_group_exists() -> bool {
```
BraveChicken1:
```suggestion
pub fn does_packit_group_exist() -> bool {
```

src/platforms/permissions/windows.rs
```diff
@@ -265,6 +265,11 @@ pub fn set_packit_permissions(path: &PathBuf, is_multiuser: bool, recurse: bool)
     }
 }
 
+/// Checks if the packit group exists.
+pub fn packit_group_exists() -> bool {
+    get_group_sid().is_ok()
```
BraveChicken1: Is using `is_ok` the correct approach here? If there is an actual error, we now think the packit group does not exist, while it could be some other issue.

src/platforms/permissions.rs
```diff
@@ -37,9 +37,12 @@ pub fn is_writable(path: &PathBuf) -> Result<bool> {
     platform::is_writable(path, metadata)
 }
 
-/// Sets the permissions of packit files
+/// Sets the permissions of packit files.
 pub use self::platform::set_packit_permissions;
 
+/// Checks if the packit group exists.
+pub use self::platform::packit_group_exists;
```
BraveChicken1: Don't forget to also implement a fallback implementation for unsupported targets

src/utils/constants.rs
```diff
@@ -5,3 +5,5 @@
 
 pub const REGISTER_FILENAME: &str = "Installed.toml";
 pub const CONFIG_FILENAME: &str = "Config.toml";
+pub const DEFAULT_METADATA_REPOSITORY_NAME: &str = "core";
```
BraveChicken1, line 10:
```suggestion
pub const CONFIG_FILENAME: &str = "Config.toml";

pub const DEFAULT_METADATA_REPOSITORY_NAME: &str = "core";
```

src/verifier/checks.rs
```diff
@@ -0,0 +1,177 @@
+use std::collections::HashSet;
+
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+pub enum Check {
+    // Initial checks (checks which verify methods which the verifier uses internally)
```
BraveChicken1:
```suggestion
    // Initial checks (checks which verify methods that are used by the verifier internally)
```

src/verifier/checks.rs
```diff
@@ -0,0 +1,177 @@
...
+                Self::RegisterExistence,
+            ],
+
+            // General checks
+            Self::PackitGroup => &[],
+            Self::StrayDirectory => &[Self::PackitGroup],
+            Self::StorageConsistency => &[Self::PackitGroup, Self::StrayDirectory],
+            Self::RegisterConsistency => &[Self::PackitGroup, Self::StrayDirectory],
+            Self::DependencyTree => &[Self::PackitGroup, Self::StorageConsistency, Self::RegisterConsistency],
+            Self::Alterations => &[Self::PackitGroup, Self::StorageConsistency, Self::RegisterConsistency],
```
BraveChicken1, line 50: Do these checks really need the PackitGroup check to be finished first?

src/verifier/checks.rs
```diff
@@ -0,0 +1,177 @@
...
+    PackageExistence,
+    PackageStorageConsistency,
+    PackageRegisterConsistency,
+    PackageDependencyTree,
+    PackageAlterations,
+}
+
+impl Check {
+    /// Gets the dependencies of a check (the checks which should happen before the given check).
+    fn get_dependencies(&self) -> &[Self] {
```
BraveChicken1: I think these functions could maybe all be `const`, the return type could then maybe even be just `[Self]`

src/verifier/error.rs
```diff
@@ -20,7 +23,13 @@ pub enum VerifierError {
     #[error("Cannot parse filename, because it contains invalid unicode")]
     InvalidUnicodeError,
 
-    #[error("Could not verify")]
+    #[error("Cannot do general checks before doing intial checks")]
```
BraveChicken1:
```suggestion
    #[error("Cannot do general checks before doing initial checks")]
```

src/verifier/error.rs
```diff
@@ -20,7 +23,13 @@ pub enum VerifierError {
...
+    InitialChecksSkippedError,
+
+    #[error("Cannot continue with checks, missing check implementation")]
+    UnimplementedCheckError,
+
+    #[error("Could not verify or fix")]
```
BraveChicken1: This error message does not seem to cover the actual error

src/verifier/issue.rs
```diff
@@ -1,85 +1,143 @@
 // SPDX-License-Identifier: GPL-3.0-only
 use colored::Colorize;
-use std::fmt::Display;
+use std::{collections::HashSet, fmt::Display, path::PathBuf};
 
 use crate::installer::types::PackageId;
 
 /// Holds a single issue and the data regarding that issue.
 pub enum Issue {
+    /// The user cannot write to the prefix directory (or one if its sub directories).
```
BraveChicken1:
```suggestion
    /// The user cannot write to the prefix directory (or one of its sub directories).
```

src/verifier/issue.rs
```diff
@@ -1,85 +1,143 @@
...
-                write!(f, "Packit group missing\n")?;
-                write!(f, "The 'packit' group is missing while multiuser mode is turned on.\n")?;
+                writeln!(f, "Packit group missing")?;
+                writeln!(f, "The 'packit' group is missing while multiuser mode is turned on.")?;
+            },
+            Issue::NotFound(package_id) => {
+                writeln!(f, "Package existence")?;
+                writeln!(f, "{} cannot be found anywhere in Packit.", package_id.to_string().bold().blue())?
+
+                // TODO: Somehow show result of fuzzy search here
```
BraveChicken1, line 108: Is this still needed?

src/verifier/repairer.rs
```diff
@@ -1,49 +1,190 @@
...
     }
 
+    /// Fixes a missing Config.toml. Either by rebuilding the config from known information or using default values.
+    fn fix_missing_config(&self) -> Result<()> {
+        // Create a default config and adjust when fields can be recovered so new config fields don't create bugs
+        let mut default_config = EditableConfig::default()?;
+
+        // Figure out the prefix path
+        let mut prefix_string = DEFAULT_PREFIX.to_string();
+        let mut prefix_path = PathBuf::from(&prefix_string);
```
BraveChicken1: You probably don't need to create a string here

src/verifier/repairer.rs
```diff
@@ -1,49 +1,190 @@
...
+                }
+
+                default_config.set_repositories_rank(new_rank);
+            }
+        } else {
+            println!("Could not open or parse '{REGISTER_FILENAME}' from '{prefix_string}', using the default repositories instead");
+        }
+
+        // Set multi-user to true if the packit group exists
+        default_config.set_multiuser(packit_group_exists());
```
BraveChicken1: Shouldn't this check if the prefix directory has the group or single user permissions?

src/verifier/repairer.rs
```diff
@@ -1,49 +1,190 @@
...
+
+        self.confirm_config_construction(&mut default_config)?;
+
+        Ok(())
```
```suggestion
        self.confirm_config_construction(&mut default_config)
```

src/verifier/verifier.rs
```diff
@@ -80,35 +148,53 @@ impl<'a> Verifier<'a> {
         }
     }
 
+    /// A wrapper around the `next_package_issue_impl` method.
```
BraveChicken1: Public functions should not include internal behaviour or private functions in its documentation. Please also include a description of what the function does here.

src/verifier/verifier.rs
```diff
@@ -225,55 +412,52 @@ impl<'a> Verifier<'a> {
...
+            if !self.directory_is_empty(&package.path())? {
+                return Ok(false);
+            }
+        }
 
-                // Get the version, and create the package id
-                let version = Version::from_str(file_version.file_name().to_str().ok_or(VerifierError::InvalidUnicodeError)?)?;
-                let package_id = PackageId::new(package_name.clone(), version);
+        Ok(true)
+    }
```
BraveChicken1: This function could be moved to the `utils::io` module

src/verifier/verifier.rs
```diff
@@ -225,55 +412,52 @@ impl<'a> Verifier<'a> {
...
 
-                // Check if the package version also exists in the register, if not add it to missing
-                if register.get_package_version(&package_id).is_none() {
-                    missing.push(package_id);
-                }
+    /// Checks if all packages in storage also exist in the register.
+    /// Returns a register consistency issue or None if there are packages missing from the register.
+    fn check_register_consistency(&self, register: &PackageRegister, config: &Config) -> Result<Option<Issue>> {
+        let storage_packages = get_storage_packages(config)?;
+        let mut missing = Vec::new();
```
BraveChicken1:
```suggestion
        let mut missing = HashSet::new();
```

BraveChicken1 (changes requested):
A lot of really nice changes!

CHANGELOG.md
Max13245: Which specific changes? I agree that it should only include behaviour relevant to the user, but some of these indirectly impact the user. For example no re-installs makes a fix a lot quicker.

README.md
Max13245: Yes, I'll put it as TODO for the refactor.

src/cli/commands/fix.rs
Max13245: It checks if the index is the same as the previously found issue (meaning the same issue is found and the fix didn't work). I might want to refactor this a bit (keeping track of the previous issue instead of it's index), but comparing issues is also annoying.

src/platforms/permissions/windows.rs
Max13245: Yes ,I'll make sure to check for the specific error.

CHANGELOG.md
BraveChicken1: I meant to comment on the "Add support for 'check dependencies'", it seems that GitHub messed up the highlighting.

src/verifier/checks.rs
Max13245: Not necessarily, but it might be better because their fixes install packages or modify metadata. If the install of a package is done with multiuser then those permissions are set wrong or fail. Then we can never get to the PackitGroup check.

src/verifier/checks.rs
Max13245: I'll make them `const`, but the size of Self cannot be known at compile time, that's why a reference is returned.

src/verifier/error.rs
Max13245: Which of these two do you mean?

I think `UnimplementedCheckError` covers the error. The IO error is just a wrapper, so the IO error itself will show the actual error.

src/verifier/repairer.rs
Max13245: I need it because otherwise I get lifetime issues. The string given by the user input doesn't life long enough.

src/verifier/repairer.rs
Max13245: Ideally yes, but that would require some additional permission functionality. So for now I decided to get with a good guess instead. Which I think is okay, since we later ask the user for confirmation.

src/verifier/issue.rs
Max13245: I think it would be nice, but I don't know how to implement it nicely.

src/verifier/error.rs
BraveChicken1: I mean the IOError, I think the message there should describe that the problem is an IO error. In other error enums we use `Error while interacting with filesystem`.

src/verifier/issue.rs
BraveChicken1: You could do the fuzzy search at issue detection time, but i'm not sure if that's still something for this PR.

src/verifier/repairer.rs
BraveChicken1: But couldn't you just directly pass the string into PathBuf::from?

src/verifier/repairer.rs
BraveChicken1: That's fine, but lets put it on the TODO board for the future.

src/verifier/issue.rs
Max13245: Alright I'll look at that for a different PR.

src/verifier/repairer.rs
Max13245: I remember now, I did it because displaying the PathBuf was difficult here. I could use to_string_lossy() and then it should be fine.

BraveChicken1 (approved):

PR #82, opened 2026-05-05
Add skip update

Max13245:
When an update cannot be done, because not all dependents of a package support the newer version. Ask the user if they want to install the newer version as active instead (keeping the older version).

### DIFF
--- a/src/installer/installer.rs
+++ b/src/installer/installer.rs
@@ -730,9 +730,15 @@ impl<'a> Installer<'a> {
             };
 
             if !dependency.satisfied(&new_package_id.name, new_version) {
-                return Err(InstallerError::SatisfyError {
-                    new_version: new_version.clone(),
-                });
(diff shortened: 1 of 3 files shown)


Max13245:
I'm a bit sad about the borrowing issue situation in the update method, but I have no idea how to fix it atm.

src/installer/installer.rs
BraveChicken1, line 771: You probably don't need the clone here, since the hashset itself already contains cloned items

BraveChicken1 (changes requested):

src/installer/installer.rs
Max13245: No I can't (sadly), because of the borrowing issues on the next for loop which also uses dependents.

BraveChicken1: Too bad, lets keep the clone then

BraveChicken1 (approved):
You're free to merge after you adjusted the CHANGELOG.

PR #83, opened 2026-05-11
Add init command

BraveChicken1:
This PR adds the init command, which can be used to initialize Packit. It sets up the config, register and prefix structure.

It currently requires the user to create the config dir and prefix dir before running the commands, and setting the correct permissions. In the future we probably want to look into this and change this behaviour.

### DIFF
--- a/src/cli/commands/init.rs
+++ b/src/cli/commands/init.rs
@@ -28,6 +28,7 @@ use crate::{
 #[derive(Args, Debug)]
 pub struct InitArgs {
     /// The prefix to use
+    #[arg(long)]
     prefix: Option<PathBuf>,
 }
 
--- a/src/cli/commands/init.rs
+++ b/src/cli/commands/init.rs
@@ -1,13 +1,21 @@
 // SPDX-License-Identifier: GPL-3.0-only
-use std::{collections::HashSet, path::PathBuf, process::exit, str::FromStr};
+use std::{
+    collections::HashSet,
+    path::{self, Path, PathBuf},
+    process::exit,
+    str::FromStr,
+};
 
 use clap::Args;
 
 use crate::{
     cli::{commands::HandleCommand, display::logging::error},
     config::{Config, EditableConfig},
-    installer::types::{PackageId, PackageName, Version},
-    platforms::DEFAULT_PREFIX,
+    installer::{
+        Symlinker,
+        types::{PackageId, PackageName, Version},
+    },
+    platforms::{DEFAULT_CONFIG_DIR, DEFAULT_PREFIX, permissions},
     repositories::types::Licenses,
     storage::{installed_package_version::InstalledPackageVersion, package_register::PackageRegister},
     utils::{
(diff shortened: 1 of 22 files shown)


src/cli/commands/init.rs
```diff
@@ -0,0 +1,143 @@
...
+        if !config_dir.exists() {
+            error!(msg: "Packit cannot be initialized: the config directory at {DEFAULT_CONFIG_DIR} does not exist yet, please create it first");
+            exit(1);
+        }
+
+        // Check if config directory is writable
+        if !permissions::is_writable(&config_dir.to_path_buf()).unwrap_or_exit_msg("Unable to check if config directory is writable", 1) {
+            error!(msg: "Packit cannot be initialized: the config directory at {DEFAULT_CONFIG_DIR} is not writable, please set the correct permissions");
+            exit(1);
+        }
```
Max13245, line 48: If we have support for specific checks in the verifier in the future. Then we could use its permission check here instead (and then we also have a fix).

src/cli/commands/init.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+impl HandleCommand for InitArgs {
+    fn handle(&self) {
+        // Check if config directory exists
+        let config_dir = Path::new(DEFAULT_CONFIG_DIR);
+        if !config_dir.exists() {
+            error!(msg: "Packit cannot be initialized: the config directory at {DEFAULT_CONFIG_DIR} does not exist yet, please create it first");
+            exit(1);
+        }
```
Max13245, line 42: Maybe offer to create it for the user automatically instead.

src/cli/commands/init.rs
Max13245: Actually, we don't have separate check for the config directory, only for the prefix directory. I'll add it as todo for the verifier.

src/cli/commands/init.rs
```diff
@@ -0,0 +1,143 @@
...
+
+        let packit_version = env!("CARGO_PKG_VERSION");
+
+        // Check if packit binary is at the correct location
+        let packit_package_path = prefix_directory.join("packages").join("packit").join(packit_version);
+        let packit_binary = packit_package_path.join("bin").join("packit");
+        if !packit_binary.exists() {
+            error!(msg: "Packit cannot be initialized: expected packit binary at {}", packit_binary.display());
+            exit(1);
+        }
```
Max13245, line 87: Do we maybe also want to check if the symlink for the "pit" command exists?

Also, it's currently in the install script, but should we maybe move the creation of the "pit" symlink to the build script instead?

src/cli/commands/init.rs
BraveChicken1: The problem is that we need sudo rights for this. We could use `Command` to spawn a sudo process, but that would not work on Windows or on Unix systems without sudo.

src/cli/commands/init.rs
BraveChicken1: We could check for the pit command too, but we assume its there, the init command mainly checks this to make sure the prefix is correct.

The pit symlink is created by the new build-install xtask. Before releasing the next version, we can change the install script to use the build-install xtask and this init command.

src/cli/commands/init.rs
Max13245: Mmhhh, alright it's good enough for now.

Max13245 (approved):

PR #84, opened 2026-05-11
Add package all and sorted flags

Max13245:
This PR adds a --sorted flag to sort packages into a prebuild directory and also a --all flag to package all install packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ Links the specified package into the /bin, /lib, /share, etc. directories. If th
 #### `pit unlink <PACKAGE-NAME>`
 Unlinks the specified package, causing the package to be unavailable from the `PATH` environment variable.
 
-#### `pit package <PACKAGE-NAME>@<VERSION> ... <DESTINATION> [--sorted] [--all]`
+#### `pit package <DESTINATION> <PACKAGE-NAME>@<VERSION> ... [--sorted] [--all]`
 Packages the specified package(s) into a prebuild and stores it in the destination directory, together with a checksum of the prebuild. When `--sorted` is used the packages will be sorted into a prebuild directory structure. `--all` will package all installed packages.
 
 #### `pit util checksum <URL>`
--- a/README.md
+++ b/README.md
@@ -84,8 +84,8 @@ Links the specified package into the /bin, /lib, /share, etc. directories. If th
 #### `pit unlink <PACKAGE-NAME>`
 Unlinks the specified package, causing the package to be unavailable from the `PATH` environment variable.
 
-#### `pit package <PACKAGE-NAME>@<VERSION> <DESTINATION>`
-Packages the specified package into a prebuild and stores it in the destination directory, together with a checksum of the prebuild.
+#### `pit package <PACKAGE-NAME>@<VERSION> ... <DESTINATION> [--sorted] [--all]`
+Packages the specified package(s) into a prebuild and stores it in the destination directory, together with a checksum of the prebuild. When `--sorted` is used the packages will be sorted into a prebuild directory structure. `--all` will package all installed packages.
 
 #### `pit util checksum <URL>`
 Calculates the checksum of the file at the given url.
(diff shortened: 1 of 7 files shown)


src/cli/commands/package.rs
```diff
@@ -10,19 +10,27 @@ use crate::{
...
 /// Packages the specified package into a prebuild and store it in the destination directory, together with a checksum of the prebuild.
 #[derive(Args, Debug)]
 pub struct PackageArgs {
-    /// The package id of the package
-    pub package_id: PackageId,
-
     /// Destination of the compressed package
     pub destination: PathBuf,
+
+    /// The package id of the package
```
BraveChicken1:
```suggestion
    /// The ids of the packages to package
```

src/cli/commands/package.rs
```diff
@@ -10,19 +10,27 @@ use crate::{
...
+    pub packages: Vec<PackageId>,
+
+    /// True to sort the package into a prebuild directory
+    #[arg(short, long, default_value = "false")]
+    pub sorted: bool,
+
+    /// True to build all installed packages
```
BraveChicken1:
```suggestion
    /// True to package all installed packages
```

src/cli/commands/package.rs
```diff
@@ -31,13 +39,40 @@ impl HandleCommand for PackageArgs {
...
+            error!(msg: "Nothing packaged, no packages specified");
+            exit(1);
+        }
+
+        for package_id in &packages {
+            // Get the correct install directory
+            let destination = match self.sorted {
+                true => {
+                    let prefix = package_id.name.get_prefix().to_string();
+                    &self.destination.join("packages").join(prefix).join(&package_id.name).join(package_id.version.to_string())
```
BraveChicken1, line 57: We should probably adjust the FilesystemPrebuildProvider and WebPrebuildProvider to use the same structure, now this would only work with the FilesystemPrebuildProvider.

src/cli/commands/package.rs
```diff
@@ -31,13 +39,40 @@ impl HandleCommand for PackageArgs {
...
+        if packages.is_empty() {
+            error!(msg: "Nothing packaged, no packages specified");
+            exit(1);
+        }
+
+        for package_id in &packages {
+            // Get the correct install directory
+            let destination = match self.sorted {
+                true => {
+                    let prefix = package_id.name.get_prefix().to_string();
```
BraveChicken1, line 56: Do you need the `to_string` here? Couldn't you directly put it in the join without it?

README.md
```diff
@@ -84,8 +84,8 @@ Links the specified package into the /bin, /lib, /share, etc. directories. If th
 #### `pit unlink <PACKAGE-NAME>`
 Unlinks the specified package, causing the package to be unavailable from the `PATH` environment variable.
 
-#### `pit package <PACKAGE-NAME>@<VERSION> <DESTINATION>`
-Packages the specified package into a prebuild and stores it in the destination directory, together with a checksum of the prebuild.
+#### `pit package <PACKAGE-NAME>@<VERSION> ... <DESTINATION> [--sorted] [--all]`
```
```suggestion
#### `pit package <DESTINATION> <PACKAGE-NAME>@<VERSION> ... [--sorted] [--all]`
```

BraveChicken1 (changes requested):

src/cli/commands/package.rs
Max13245: Nope, the `to_string` method is necessary to pass it as a parameter to `join`.

src/cli/commands/package.rs
Max13245: Yes, good point. I actually think we should go with the structure from the FilesystemPrebuildProvider, because the structure from the WebPrebuildProvider doesn't account for different OS versions well. If we go with the FilesystemPrebuildProvider structure we can easily have multiple OS version targets next to each other when needed.

BraveChicken1 (approved):

PR #85, opened 2026-05-12
Refactor fuzzy alternatives

Max13245:
This PR is a refactor for all the fuzzy alternative display code. 
- It's now in a separate module, making the command code easier to read.
- The info command now also has version suggestions.

### DIFF
--- a/src/cli/commands/update.rs
+++ b/src/cli/commands/update.rs
@@ -4,7 +4,10 @@ use std::process::exit;
 use clap::Args;
 
 use crate::{
-    cli::{commands::HandleCommand, display::logging::error},
+    cli::{
+        commands::HandleCommand,
+        display::{logging::error, not_found},
+    },
     config::Config,
     installer::{
         Installer, InstallerOptions,
@@ -13,7 +16,7 @@ use crate::{
     platforms::Target,
     repositories::manager::RepositoryManager,
     storage::package_register::PackageRegister,
-    utils::{fuzzy, unwrap_or_exit::UnwrapOrExit},
+    utils::unwrap_or_exit::UnwrapOrExit,
 };
 
 /// Updates the specified package to the new version, or the latest version if no new version is specified.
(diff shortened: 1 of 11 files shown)


src/cli/commands/update.rs
```diff
@@ -34,32 +37,11 @@ impl HandleCommand for UpdateArgs {
...
-                println!("Did you mean: '{fuzzy_match}'?");
-            }
-
-            return;
+        match self.optional_id.versioned() {
+            Some(package_id) if register.get_package_version(&package_id).is_some() => {},
+            Some(package_id) => not_found::register_package_version(&package_id, &register),
+            None if register.get_package(&self.optional_id.name).is_some() => {},
+            None => not_found::register_package(&self.optional_id.name, &register),
         }
```
BraveChicken1, line 45: This seems to be used in multiple places, maybe it's worth also moving this to the `not_found` module.

src/cli/display/not_found.rs
```diff
@@ -0,0 +1,92 @@
...
+    cli::display::logging::error,
+    config::Config,
+    installer::types::{PackageId, PackageName, Version},
+    repositories::manager::RepositoryManager,
+    storage::package_register::PackageRegister,
+    utils::{fuzzy, unwrap_or_exit::UnwrapOrExit},
+};
+
+/// Shows possible register versions and exits with status 1 if the given package name exits.
+pub fn register_version(package_name: &PackageName, register: &PackageRegister) {
```
BraveChicken1, line 14:
```suggestion
pub fn register_version(package_name: &PackageName, register: &PackageRegister) -> ! {
```

src/cli/display/not_found.rs
```diff
@@ -0,0 +1,92 @@
...
+    // Return early if the package name doesn't exist in the register
+    let Some(package) = register.get_package(package_name) else {
+        return;
+    };
+
+    display_versions(package.versions.keys())
+}
+
+/// Shows possible repository versions and exits with status 1 if the given package name exits.
+pub fn repository_version(package_name: &PackageName, manager: &RepositoryManager) {
```
BraveChicken1, line 24:
```suggestion
pub fn repository_version(package_name: &PackageName, manager: &RepositoryManager) -> ! {
```

BraveChicken1 (changes requested):

src/cli/display/not_found.rs
Max13245: This method doesn't always exit, so the return type should not be "!".

src/cli/display/not_found.rs
Max13245: This method doesn't always exit, so the return type should not be "!".

src/cli/commands/update.rs
Max13245: I think it's nice to do the checks for package existence in the command themselves. Also, 2 of the 3 use the register the other one uses the repository.

BraveChicken1 (approved):

PR #86, opened 2026-05-13
Add checks and fixes for register fields

Max13245:
This PR adds checks and fixes for the register fields.

This does NOT yet fix the broken dependency tree bug.

### DIFF
--- a/src/verifier/repairer.rs
+++ b/src/verifier/repairer.rs
@@ -61,6 +61,7 @@ impl Repairer {
             Issue::InvalidDependents(invalid) => self.fix_invalid_dependents(invalid, register),
             Issue::InvalidActive(invalid) => self.fix_invalid_active(invalid, register, config)?,
             Issue::ForbiddenLink(forbidden) => self.fix_forbidden_link(forbidden, register, config)?,
+            Issue::MissingLinks(missing) => self.fix_missing_links(missing, register, config)?,
             _ => warning!("Fix not executed, because the issue fix is not yet implemented"),
         }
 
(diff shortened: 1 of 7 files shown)


src/verifier/repairer.rs
```diff
@@ -278,12 +286,14 @@ impl Repairer {
...
-                &Target::current(),
-            )?;
+            let prebuild_url = manager
+                .get_prebuild_url(
+                    &repository_id,
+                    package_id,
+                    package_version_meta.revisions.len() as u64,
+                    &Target::current(),
+                )
+                .unwrap_or(None);
```
BraveChicken1: I think we shouldn't try to set a prebuild_url, since we don't know if it was actually installed from a prebuild.

src/verifier/repairer.rs
```diff
@@ -355,4 +365,142 @@ impl Repairer {
...
+
+            // Set the active to the latest installed version of the package
+            if let Some(version) = package.versions.keys().into_iter().max() {
+                package.active_version = version.clone();
+            }
+
+            // Remove symlink for the package name if it exists
+            let active_symlink = active_directory.join(&package_name);
+            if fs::exists(&active_symlink)? {
+                remove_symlink(&active_symlink)?;
```
```suggestion
                symlink::remove_symlink(&active_symlink)?;
```

src/verifier/repairer.rs
```diff
@@ -355,4 +365,142 @@ impl Repairer {
...
+            }
+
+            // Link the correct active version
+            let active_original = package_directory.join(&package_name).join(package.active_version.to_string());
+            create_symlink(&active_original, &active_symlink)?;
```
```suggestion
            symlink::create_symlink(&active_original, &active_symlink)?;
```

src/verifier/repairer.rs
```diff
@@ -355,4 +365,142 @@ impl Repairer {
...
+                warning!("Could not fix invalid dependents for {child}");
+                continue;
+            };
+
+            package_version.dependents.remove(&parent);
+        }
+    }
+
+    /// Fixes the invalid active issue.
+    fn fix_invalid_active(&self, invalid: Vec<PackageName>, register: &mut PackageRegister, config: &Config) -> Result<()> {
```
BraveChicken1, line 443: Could you not use the Symlinker::set_active function here?

src/verifier/verifier.rs
```diff
@@ -576,6 +994,34 @@ impl Verifier {
...
+        let mut package_version_meta = None;
+        for repository_id in &config.repositories_rank {
+            let Some(repository) = config.repositories.get(repository_id) else {
+                continue;
+            };
+
+            // Continue if the repository doesn't match
+            if repository.path != package.source_repository_url {
+                continue;
+            }
```
BraveChicken1: You should create a repository from the source_repository_url and source_repository_provider from the InstalledPackageVersion. The config should not be used here.

BraveChicken1 (changes requested):

src/verifier/repairer.rs
Max13245: I think we should. Eventually I want to expand this with checksums, which would make sure we got the correct repository. The extra information can't hurt in this case.

BraveChicken1 (approved):

PR #89, opened 2026-05-14
General refactor

BraveChicken1:
This PR does a general refactor of Packit:
- Some small refactorings to improve code quality and readability
- Add packit_version module
- Rename storage module to register, rename Installed.toml to Register.toml and rename source_repository fields
- Split the builder, build_env and binary patcher into a new builder module
- Split the util command into separate modules
- Change WebPrebuildProvider structure
- Rename sorted flag of packager to structured
- Fix warnings and clippy lints


src/repositories/manager.rs
```suggestion
match self.metadata_providers.get(repository_id) {
            Some(provider) => provider.read_file_bytes(package, file_path),
            None => Err(RepositoryError::RepositoryNotFoundError {
                repository_id: repository_id.into(),
            }),
        }
```

src/installer/installer.rs
Max13245: Make clippy happy (Path is not owned):
```suggestion
        if self.config.prefix_directory != Path::new(DEFAULT_PREFIX) {
```

src/cli/commands/util.rs
Max13245, line 1: Didn't we also want to put the `pit package` command in the utils?

src/builder/error.rs
```diff
@@ -47,6 +44,9 @@ pub enum BuilderError {
     #[error("Cannot request files for building")]
     RequestError(#[from] reqwest::Error),
 
+    #[error("Build files download unsuccessful, with status code: {0}.")]
+    RequestUnsuccessful(reqwest::StatusCode),
+
```
Max13245: Why is this moved here? Generally we have all the #[from] stuff at the bottom.

src/installer/installer.rs
```diff
@@ -646,7 +646,7 @@ impl<'a> Installer<'a> {
         let provider = match provider::create_metadata_provider(repository) {
             Some(provider) => provider,
             None => {
-                error!(msg: "Unable to create repository provider to retrieve uninstall script");
+                error!(msg: "Unable to create repository provider, skipping install script execution. This may cause stray files");
```
Max13245: Shouldn't this be:
Unable to create repository provider, skipping uninstall script execution. This may cause stray files.

src/installer/installer.rs
```diff
@@ -655,7 +655,10 @@ impl<'a> Installer<'a> {
         let package_version = match provider.read_package_version(&package_id.name, &package_id.version) {
             Ok(package_version) => package_version,
             Err(e) => {
-                error!(e, "Unable to read package version from source repository");
+                error!(
+                    e,
+                    "Unable to read package version from source repository, skipping install script execution. This may cause stray files"
```
Max13245: Same thing, shouldn't this be:
Unable to read package version from source repository, skipping uninstall script execution. This may cause stray files.

src/installer/scripts.rs
```diff
@@ -45,6 +33,18 @@ pub enum ScriptError {
...
+    RunError(std::io::Error),
+
+    #[error("Cannot transform path to absolute path: {0}")]
+    AbsolutePathError(std::io::Error),
+
+    #[error("Cannot save script to file: {0}")]
+    SaveError(std::io::Error),
+
+    #[error("Cannot create temp directory for script to run: {0}")]
+    TempCreationError(std::io::Error),
```
Max13245: Same question, why did we move this to the bottom?

src/register/package_register.rs
```diff
@@ -39,17 +43,17 @@ impl PackageRegister {
...
+        // If the file does not exist, return an error
         if !fs::exists(path)? {
-            return Ok(Self { packages: HashMap::new() });
+            return Err(RegisterError::RegisterDoesNotExist);
         }
 
         // Read data from file
         let file_content = fs::read_to_string(path)?;
 
         // If the file is empty, return an empty storage
```
Max13245: Also adjust the comment here to:
// If the file is empty, return an error

src/installer/installer.rs
Max13245, line 246:
```suggestion
// Get metadata repository for installed storage before actually installing package
let metadata_repository = self.config.repositories.get(&install_meta.repository_id).expect("Expected repository in config");
```

src/installer/installer.rs
```diff
@@ -521,7 +521,10 @@ impl<'a> Installer<'a> {
 
         // Load source repository
```
Max13245:
```suggestion
        // Load repository
```

src/installer/installer.rs
```diff
@@ -613,7 +616,10 @@ impl<'a> Installer<'a> {
         // Run uninstall scripts for all versions
         for package_version in &installed_versions {
             // Load source repository
```
Max13245:
```suggestion
            // Create repository
```

src/installer/installer.rs
Max13245: Also:

```suggestion
                    "Unable to read package version from metadata repository, skipping install script execution. This may cause stray files"
```

src/installer/installer.rs
Max13245:
```suggestion
        // Create repository provider for metadata repository
```

Max13245:
```suggestion
        // Load package version from metadata repository
```

src/register/package_register.rs
Max13245, line 100:
```suggestion
        repository: &Repository,
```

src/verifier/repairer.rs
Max13245, line 322:
```suggestion
            let repository = config.repositories.get(&repository_id).expect("Expected repository in config");
```

CHANGELOG.md
```diff
@@ -10,24 +10,22 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
 - The fuzzy-search feature, to provide a package suggestion in case of a wrong package parameter.
-- Add support for leading zeros in version numbers.
-- Add web prebuild provider for web prebuild support.
+- Support for leading zeros in version numbers.
+- The web prebuild provider for web prebuild support.
 - The `--updatables` flag on the list command, to list packages which can be updated.
-- The list command now has grid printing.
-- The regex search, to search packages based on a given regex pattern.
-- Add support for applying patches to the source code of packages.
+- Regex search to the search command, to search packages based on a given regex pattern.
```
Max13245: Maybe this instead:
- Regex search, to search packages based on a given regex pattern.

Or:
- Add regex search to the search command, to search packages based on a given regex pattern.

src/installer/installer.rs
Max13245: Another thing, in my opinion this should be a warning. Otherwise the user gets this:
`pit uninstall some-package`
Error: Unable to read package version from source repository, skipping install script execution. This may cause stray files.

This is very confusing. Did the uninstall work? Did it not? A warning would indicate something went wrong, but in general everything is fine.

Max13245 (changes requested):
Overal it looks good. Just two things. 

Do we look at the install scripts after the release?

It might be nice to also rename path -> metadata_url and metadata_provider (of meta_url en meta_provider) in the Repository struct. The url part wil make it consistent with the provider_url. I think the metadata part is also nice, as it makes a clear indication of what the url is for.

src/installer/installer.rs
Max13245: So, just in total:
warning!("Unable to read package version from metadata repository, skipping uninstall script execution. This may cause stray files.")

src/installer/installer.rs
BraveChicken1: This is unfortunately not allowed, the other option would be:
`if self.config.prefix_directory != *DEFAULT_PREFIX {`

BraveChicken1: The Repository struct also contains prebuild repository information, so here we should not use metadata_repository.

src/register/package_register.rs
BraveChicken1: I'm not sure about this one, don't we want to specify that this is the repository that we installed the package from (hence source_repository)

src/verifier/repairer.rs
BraveChicken1: Same comment for this one, do we want to rename all occurrences of source_repository to just repository?

src/installer/scripts.rs
BraveChicken1: The idea was to put all external errors to the bottom, but I moved it back up.

src/builder/error.rs
BraveChicken1: Yes, that shouldn't have been moved, I moved it back up

src/installer/installer.rs
Max13245: I tested this locally. For me it works? self.config.prefix_directory is a PathBuf, so it can be compared to the Path right? Although *DEFAULT_PREFIX is also fine.

BraveChicken1: I tried it again, it does indeed work for me now, I changed it.

Max13245 (approved):

PR #91, opened 2026-05-21
Add package-build xtask

BraveChicken1:
This PR adds the package-build xtask, packaging the most recent build into a prebuild. For compressing the existing packager module is copied and minified to use less dependencies and be compressed in the correct structure as expected by the install scripts.

### DIFF
--- a/xtasks/xtask-package-build/src/main.rs
+++ b/xtasks/xtask-package-build/src/main.rs
@@ -1,0 +1,93 @@
+// SPDX-License-Identifier: GPL-3.0-only
+mod compress;
+mod target_architecture;
+
+use std::{
+    fs,
+    path::{Path, PathBuf},
+    process::exit,
+};
+
+use clap::Parser;
+
+use crate::target_architecture::TARGET_ARCHITECTURE;
+
+/// Represents the package-build command, holding the given options.
+#[derive(Parser, Debug)]
+#[command(name = "package-build", version, about)]
+pub struct TaskCommand {
+    /// The directory to save the package to
+    #[arg(long)]
+    destination: Option<PathBuf>,
+
+    /// True if the package should be overwritten if it already exists
+    #[arg(short, long, default_value = "false")]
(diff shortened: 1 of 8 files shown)


xtasks/xtask-package-build/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+    /// The directory to save the package to
+    #[arg(long)]
+    destination: Option<PathBuf>,
+
+    /// True if the package should be overwritten if it already exists
+    #[arg(short, long, default_value = "false")]
+    overwrite: bool,
+}
+
+const PACKIT_VERSION: &str = "0.0.2";
```
Max13245, line 28: Is it possible to get the version from cargo or is it not possible to get this information from another cargo env?

xtasks/xtask-package-build/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+            exit(1);
+        },
+    };
+
+    // Check if package already exists
+    let package_name = format!("packit@{PACKIT_VERSION}-{PACKIT_REVISION}-{TARGET_ARCHITECTURE}");
+    let package_file_name = format!("{package_name}.tar.gz");
+    let archive_destination = destination.join(&package_file_name);
+    if !command.overwrite && archive_destination.exists() {
+        eprintln!("The package already exists, please remove it first or specifiy the -o option");
```
Max13245: Maybe use the long format for the flag to specify its intend:
```suggestion
        eprintln!("The package already exists, please remove it first or specifiy the `--overwrite` option");
```

xtasks/xtask-package-build/src/main.rs
BraveChicken1: I tried this, unfortunately there does not seem to be a good and reliable way to do this. The most reliable option would be to parse the Cargo.toml file, but I don't think that is really desirable.

Max13245 (approved):

PR #92, opened 2026-05-23
Remove unnecessary clones

Max13245:
This PR removes some unnecessary clones.

### DIFF
--- a/src/installer/installer.rs
+++ b/src/installer/installer.rs
@@ -71,11 +71,13 @@ impl<'a> Installer<'a> {
         let (repository_id, package_metadata) = self.repository_manager.read_package(&optional_id.name)?;
         let latest_version = package_metadata.get_latest_version(&Target::current())?;
 
+        // Create a package id of the current package
+        let package_id = optional_id.versioned_or(latest_version);
+
         // If the version isn't specified check if a package with this package name is already installed (otherwise a user can get two different version installed without knowing)
         if optional_id.version.is_none() {
             if let Some(package) = self.register.get_package(&optional_id.name) {
                 if package.get_package_version(latest_version).is_some() {
-                    let package_id = optional_id.versioned_or(latest_version.clone());
                     return Err(InstallerError::AlreadyInstalledError { package_id });
                 }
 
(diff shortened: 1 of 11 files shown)


src/installer/types/optional_id.rs
```diff
@@ -65,10 +65,10 @@ impl OptionalPackageId {
     }
 
     /// Returns a `PackageId` with the current version, or the given version if the `OptionalPackageId` does not contain a version.
-    pub fn versioned_or(&self, version: Version) -> PackageId {
+    pub fn versioned_or(&self, version: &Version) -> PackageId {
         let version = match &self.version {
             Some(version) => version.clone(),
-            None => version,
+            None => version.clone(),
```
BraveChicken1, line 82: I would not do this clone here. When we would have an owned version, we would now need to clone it too.

src/installer/installer.rs
```diff
@@ -712,7 +711,7 @@ impl<'a> Installer<'a> {
         // Check if the new version is lower then the current
         if old_package.package_id.version > *new_version {
             return Err(InstallerError::VersionTooLowError {
-                new_version: old_package.package_id.version.clone(),
+                new_version: new_version.clone(),
```
BraveChicken1, line 714: Is this a bugfix?

BraveChicken1 (changes requested):

src/installer/types/optional_id.rs
Max13245: No, I checked, we never have an owned version and we probably never will (because most of the time it will be a reference, since we get it from the register or manager). Also, it's slightly less efficient at the moment, because we are already creating a Version even though we are not sure yet if we will need one.

src/installer/installer.rs
Max13245: Yes, I should probably have clarified this, but this is indeed a small bugfix.

src/installer/types/optional_id.rs
BraveChicken1: Then I would create a seperate function for this, either a `_cloned` with this functionality or `_else` with a closure. We really shouldn't hide a clone in a function this way.

Max13245: Sure, the original method will be unused though.

BraveChicken1 (approved):

PR #93, opened 2026-05-25
Fix multiple repositories issues

BraveChicken1:
This PR fixes issues regarding the use of multiple repositories. When searching or installing a version which is not in the repository that is first in the rank, the next repository is checked now. This ensures the package is recognized succesfully.

### DIFF
--- a/src/cli/commands/search.rs
+++ b/src/cli/commands/search.rs
@@ -98,7 +98,7 @@ impl SearchArgs {
         };
 
         // Create a package id
-        let package_id = optional_id.versioned_or(package_version.version.clone());
+        let package_id = optional_id.versioned_or_cloned(&package_version.version);
 
         let target_bounds = package_version
             .get_best_target(&Target::current())
--- a/src/cli/commands/search.rs
+++ b/src/cli/commands/search.rs
@@ -75,7 +75,7 @@ impl SearchArgs {
 
         let config = Config::from(&Config::get_default_path()).unwrap_or_exit_msg("Cannot load config", 1);
         let manager = RepositoryManager::new(&config);
-        let (repository_id, package) = match manager.read_package(&optional_id.name) {
+        let (repository_id, package, package_version) = match manager.read_package_and_version(&optional_id, &Target::current()) {
             Ok(package) => package,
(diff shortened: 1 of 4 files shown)


src/installer/installer.rs
```diff
@@ -68,21 +68,24 @@ impl<'a> Installer<'a> {
...
-        let (repository_id, package_metadata) = self.repository_manager.read_package(&optional_id.name)?;
-        let latest_version = package_metadata.get_latest_version(&Target::current())?;
+        // Read package and version metadata
+        let (repository_id, package_metadata, version_metadata) =
+            self.repository_manager.read_package_and_version(&optional_id, &Target::current())?;
 
         // Create a package id of the current package
-        let package_id = optional_id.versioned_or_cloned(latest_version);
+        let version = &version_metadata.version;
+        let package_id = optional_id.versioned_or(version.clone());
```
Max13245: Here you could use `version_or_cloned` I think, then you only clone the value if necessary.

src/installer/installer.rs
Max13245:
```suggestion
if optional_id.version.is_none() && self.register.get_package(&optional_id.name).is_some() {
            let question = format!(
                "The package '{optional_id}' is already installed, but a newer version '{version}' is available. Do you wish to install the latest version as well?"
            );

            if ask_user(&question, QuestionResponse::No)?.is_no_or_invalid() {
                return Err(InstallerError::InstallationCanceled {
                    reason: "A version of this package was already installed".to_string(),
                });
            }
        }
```

src/cli/commands/search.rs
```diff
@@ -98,16 +98,7 @@ impl SearchArgs {
...
-
-        // Get package version info for its target
-        let package_version = match manager.read_repo_package_version(&repository_id, &package_id) {
-            Ok(package_version) => package_version,
-            Err(e) => {
-                error!(e, "Cannot read '{package_id}' from repository {repository_id}");
-                return;
-            },
-        };
+        let package_id = optional_id.versioned_or(package_version.version.clone());
```
Max13245:
```suggestion
        let package_id = optional_id.versioned_or_cloned(package_version.version);
```

Max13245 (changes requested):

Max13245 (approved):

PR #99, opened 2026-06-19
Extend patch implementation to support context and unified diffs

BraveChicken1:
This PR extends the patch implementation to support applying context diffs and unified diffs.
It also fixes the bug that patch files stored in the metadata repository were not verified with the checksum.
A better file path resolution algorithm is implemented to support a wider range of patches.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ url = "2.5.8"
 toml_edit = "0.25.11"
 terminal_size = "0.4.4"
 diffy = { version = "0.5.0", features = ["binary"] }
+contextdiff-parser = { path = "../contextdiff-parser" }
 
 # macOS and Linux dependencies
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -39,6 +39,7 @@ url = "2.5.8"
 toml_edit = "0.25.11"
 terminal_size = "0.4.4"
 diffy = { version = "0.5.0", features = ["binary"] }
+contextdiff-parser = { git = "https://github.com/pack-it/contextdiff-parser.git" }
```
Max13245, line 42: Even though it's from our own repo, maybe also lock the version here.

Max13245 (approved):

Cargo.toml
BraveChicken1: My plan was to do this when we have an official release of the contextdiff-parser crate. This should be resolved before the 0.0.3 release of Packit.

PR #100, opened 2026-06-20
Refactor tree

Max13245:
This refactors the tree util, making it easier to use (less generic types). It also makes the install iterative instead of recursive in preparation for parallelization of the install.

The install tree is also fully expanded and shown to the user, making it clearer what will be installed.

The UI functionality still needs some works (also deciding which crates to finally use), but this should be done on a separate UI branch I think.

### DIFF
--- a/src/installer/install_tree.rs
+++ b/src/installer/install_tree.rs
@@ -45,8 +45,8 @@ impl InstallLabel {
         &self.install_type
     }
 
-    /// Returns true if it is a build dependency
-    /// TODO: Change naming to is_build_dependency
+    /// Returns true if it is a normal dependency (not a build dependency).
+    /// If it is a dependency of a build dependency this also returns true.
     pub fn is_dependency(&self) -> bool {
         self.is_dependency
     }
@@ -78,14 +78,13 @@ impl InstallMeta {
 pub type InstallTree = Tree<Option<InstallMeta>, InstallLabel>;
 pub type InstallNode = Node<Option<InstallMeta>, InstallLabel>;
 
-// TODO: Maybe InstallTreeBuilder
-pub struct DependencyResolver<'a> {
+pub struct InstallTreeBuilder<'a> {
     register: &'a PackageRegister,
     repository_manager: &'a RepositoryManager<'a>,
     asked_packages: HashSet<PackageId>,
 }
 
-impl<'a> DependencyResolver<'a> {
+impl<'a> InstallTreeBuilder<'a> {
     pub fn new(register: &'a PackageRegister, manager: &'a RepositoryManager) -> Self {
         Self {
             register,
@@ -115,19 +114,23 @@ impl<'a> DependencyResolver<'a> {
                         &self.register.get_package_version(node.get_package_id()).expect("Expected package version to exist").dependencies;
                     for dependency in dependencies {
                         let new_node = Node::new(dependency.clone(), None, InstallLabel::new(InstallType::Installed, false));
-                        let new_index = tree.add_node(node_index, new_node);
+                        let new_index = tree.add_node(node_index, new_node)?;
                         package_queue.push_back(new_index);
                     }
 
                     continue;
                 },
-                None => continue, // TODO: Error
+                None => {
+                    return Err(InstallerError::UnreachableError {
+                        msg: "Node value is None without InstallType::Installed".to_string(),
+                    });
+                },
             };
 
-            for (dependency, label) in DependencyResolver::expander(node, install_meta)? {
+            for (dependency, label) in InstallTreeBuilder::expander(node, install_meta)? {
                 let (dependency_id, meta, dependency_label) = self.populator(&dependency, label)?;
                 let new_node = Node::new(dependency_id, meta, dependency_label);
-                let new_index = tree.add_node(node_index, new_node);
+                let new_index = tree.add_node(node_index, new_node)?;
                 package_queue.push_back(new_index);
             }
         }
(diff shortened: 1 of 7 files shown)


src/installer/install_tree.rs
```diff
@@ -66,26 +77,94 @@ impl InstallMeta {
...
+pub struct InstallTreeBuilder<'a> {
+    register: &'a PackageRegister,
+    repository_manager: &'a RepositoryManager<'a>,
+    checked_packages: HashSet<PackageId>,
+    asked_packages: HashSet<PackageId>,
+    terminal: Term,
+}
+
+impl<'a> InstallTreeBuilder<'a> {
+    pub fn new(register: &'a PackageRegister, manager: &'a RepositoryManager) -> Self {
```
BraveChicken1:
```suggestion
    pub fn new(register: &'a PackageRegister, repository_manager: &'a RepositoryManager) -> Self {
```

src/installer/install_tree.rs
```diff
@@ -66,26 +77,94 @@ impl InstallMeta {
...
+        let mut package_queue = VecDeque::from([0 as usize]);
+        while let Some(node_index) = package_queue.pop_front() {
+            let node = tree.get_node_by_index_mut(node_index).expect("Expected node to exist");
+
+            // Expand with register if the node value is None and has the Installed label type (meaning that the package is already installed)
+            let install_meta = match node.get_value() {
+                Some(install_meta) => install_meta,
+                None if node.get_label().install_type == InstallType::Installed => {
+                    let dependencies =
+                        &self.register.get_package_version(node.get_package_id()).expect("Expected package version to exist").dependencies;
```
BraveChicken1, line 123: Wouldn't it be cleaner to use an UnreachableError here instead of the expects?

src/installer/install_tree.rs
```diff
@@ -66,26 +77,94 @@ impl InstallMeta {
...
+                    continue;
+                },
+                None => {
+                    return Err(InstallerError::UnreachableError {
+                        msg: "Node value is None without InstallType::Installed".to_string(),
+                    });
+                },
+            };
+
+            for (dependency, label) in InstallTreeBuilder::expander(node, install_meta)? {
```
BraveChicken1: Why isn't the expander called like self.expander?

src/installer/install_tree.rs
```diff
@@ -113,36 +192,79 @@ impl InstallNode {
...
 
-    /// Expands an install tree after it has been build. The current node will be changed to a build node
-    /// and its children will be updated accordingly.
-    pub fn expand_with_build(&mut self, register: &PackageRegister, manager: &RepositoryManager) -> tree::Result<()> {
-        self.set_label(InstallLabel {
-            install_type: InstallType::Build,
-            is_dependency: self.get_label().is_dependency(),
-        });
+    fn check_prebuild(&mut self, install_meta: &InstallMeta, package_id: &PackageId, label: &InstallLabel) -> Result<Option<InstallLabel>> {
+        // Don't check for prebuild if the package should not use a prebuild or if it was already asked
```
BraveChicken1: This `if` does not do the 'already asked' check right?

src/utils/tree.rs
```diff
@@ -1,154 +1,139 @@
...
+            return Err(TreeError::CycleError(node.get_package_id().clone()));
         }
-    }
 
-    pub fn expander(mut self, expander: E) -> Self {
-        self.expander = Some(expander);
-        self
-    }
+        node.parent_index = parent_index;
+        self.nodes.push(node);
```
BraveChicken1: I would do the parent check first, to prevent an invalid tree state

src/utils/tree.rs
```diff
@@ -1,154 +1,139 @@
...
+    pub fn get_nodes(&self) -> &Vec<Node<V, L>> {
+        &self.nodes
+    }
 
-        let populator = match &self.populator {
-            Some(populator) => populator,
-            None => return Err(TreeError::MissingBuildAttributes("populator".to_string())),
-        };
+    /// Gets a node based on its index.
+    pub fn get_node_by_index(&self, index: &usize) -> Option<&Node<V, L>> {
```
```suggestion
    pub fn get_node_by_index(&self, index: usize) -> Option<&Node<V, L>> {
```

src/utils/tree.rs
```diff
@@ -1,154 +1,139 @@
...
+        node.get_children()
+            .iter()
+            .map(|c| self.get_node_by_index(c).expect("Expected node to exist"))
+            .filter(|c| filter(&c.label))
+            .map(|c| c.package_id.clone())
+            .collect()
+    }
+
+    /// Checks if a given package id will create a cycle in the tree.
+    /// Returns true is a cycle will be formed, false if not.
```
```suggestion
    /// Returns true if a cycle will be formed, false if not.
```

src/utils/tree.rs
```diff
@@ -1,154 +1,139 @@
...
         }
 
-        Ok(())
+        Ok(false)
     }
 
     /// The implementation of the tree display.
-    fn display_impl(&self, f: &mut std::fmt::Formatter<'_>, node: &Node<V, L>, prefix: &str) -> std::fmt::Result {
-        writeln!(f, "{prefix}{}", node.id)?;
+    fn display_impl(&self, f: &mut std::fmt::Formatter<'_>, node_index: &usize, prefix: &str) -> std::fmt::Result {
```
```suggestion
    fn display_impl(&self, f: &mut std::fmt::Formatter<'_>, node_index: usize, prefix: &str) -> std::fmt::Result {
```

BraveChicken1 (changes requested):

src/installer/install_tree.rs
Max13245: Mhm, maybe. Although then we would have to do it for all expects. Why is this one different?

BraveChicken1: When I read this code, I don't immediately see why this expect is allowed here. So I think we should either use an UnreachableError to indicate what has gone wrong, or add a comment justifying the expect usage here.

Max13245: I thought it was quite clear, because of the InstallType::Installed, but I'll put a comment there then.

BraveChicken1 (approved):

PR #101, opened 2026-06-21
Bump dependencies

BraveChicken1:
This PR bumps all dependencies of Packit and disables unnecessary features of dependencies.

Dependency zip is currently at a version that has a MSRV of 1.88, which is above or minimum supported. We should probably downgrade zip to 7.2.0, which has a MSRV of 1.83.


Max13245:
Yes, downgrading zip is probably the right call then. Maybe you can already do that on this branch?

Max13245 (approved):

PR #103, opened 2026-06-22
Refactor IO errors

BraveChicken1:
This PR refactors the IO errors and implements a new wrapper error type that includes information about the operation (type, path).


src/installer/installer.rs
```diff
@@ -177,8 +177,8 @@ impl<'a> Installer<'a> {
         }
 
         // Create install directory if it does not exist
-        if !fs::exists(&install_directory)? {
-            fs::create_dir_all(&install_directory)?;
+        if !fs::exists(&install_directory).err_with_path("check existence of", &install_directory)? {
+            fs::create_dir_all(&install_directory).err_with_path("create", &install_directory)?;
```
Max13245: I only commented it here, but there are a couple other places where this could be done (just search for "create" in the changes):
```suggestion
            fs::create_dir_all(&install_directory).err_with_path("create all", &install_directory)?;
```

src/installer/scripts.rs
```diff
@@ -32,17 +32,17 @@ pub enum ScriptError {
...
 
-    #[error("Cannot save script to file: {0}")]
-    SaveError(std::io::Error),
+    #[error("Cannot save script or test file to file")]
+    SaveError(#[source] std::io::Error),
 
-    #[error("Cannot create temp directory for script to run: {0}")]
-    TempCreationError(std::io::Error),
+    #[error("Cannot create temp directory for script to run")]
+    TempCreationError(#[source] std::io::Error),
```
Max13245, line 45: I'm not sure about this, do we want to add extra context in this case?

In other places for example we don't have a save error (then we also just use `err_with_path` with "write" as operation).

Max13245 (changes requested):

src/installer/installer.rs
BraveChicken1: I'm not sure if I like the operation description "create all", because the error message would then become: "Failed to create all /some/path"

src/installer/scripts.rs
BraveChicken1: Yes, but in the packager we use the same approach. I think it is neat to split it here, since scripts are a big part of the build and install process.

src/installer/installer.rs
Max13245: Mhmm, maybe "create dir all", like the command then? This is a small issue though, so if not this, you can resolve this comment.

src/installer/scripts.rs
Max13245: Mhmm, sure, I guess more context can't hurt.

Max13245 (approved):

PR #104, opened 2026-06-22
Small install fixes/changes

BraveChicken1:
The symlinker now doesn't create empty directories anymore, it first checks if the directory is needed.
The symlinker now symlinks the `gnubin` directory.
The build environment now uses an absolute path to the M4 installation in the `M4` environment variable when M4 is listed as a build dependency.

### DIFF
--- a/src/builder/build_env.rs
+++ b/src/builder/build_env.rs
@@ -57,9 +57,8 @@ impl<'a> Into<Environment> for BuildEnv<'a> {
         }
 
         // Add M4 variable if m4 is a dependency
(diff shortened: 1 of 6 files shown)


src/builder/build_env.rs
```diff
@@ -57,9 +57,8 @@ impl<'a> Into<Environment> for BuildEnv<'a> {
         }
 
         // Add M4 variable if m4 is a dependency
```
Max13245:
```suggestion
        // Add M4 variable if m4 is a build dependency
```

Max13245 (approved):

PR #105, opened 2026-06-26
Allow optional ids as parameters to the check and fix commands

Max13245:
This PR adds the ability to use optional package ids as parameters to the check and fix commands. 

The duplicates util is also renamed to `parameter_checks.rs` and the `expand_optional_ids` is placed in there. It's nice to have this in the utils, but at the same time I'm not sure how I feel about a method which exists in the utils.


src/cli/parameter_checks.rs
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        let installed_directory = config.prefix_directory.join("packages").join(&optional_id.name);
+        if register.get_package(&optional_id.name).is_none() && !fs::exists(installed_directory).unwrap_or_exit(1) {
+            not_found::register_package(&optional_id.name, &register);
+        }
+
+        let package = register.get_package(&optional_id.name).unwrap();
+        package.versions.keys().for_each(|v| package_ids.push(PackageId::new(optional_id.name.clone(), v.clone())));
+    }
```
BraveChicken1, line 58: Since this function calls the not_found module from the cli module, I would move this to the cli module too.

src/cli/parameter_checks.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/// Expands all the optional ids to package ids. Will exit with a not found error if the given optional id doesn't exist in the register.
+/// Returns a list of `PackageId`.
+pub fn expand_optional_ids(register: &PackageRegister, config: &Config, packages: &Vec<OptionalPackageId>) -> Vec<PackageId> {
+    let mut package_ids = Vec::new();
+    for optional_id in packages {
+        if let Some(package_id) = optional_id.versioned() {
+            let installed_directory = config.prefix_directory.join("packages").join(&package_id.name).join(package_id.version.to_string());
+            if register.get_package_version(&package_id).is_none() && !fs::exists(installed_directory).unwrap_or_exit(1) {
```
BraveChicken1, line 43: Because of this directory check it even seems that it should maybe be a part of the integrity module. Maybe you can move the expansion of optional ids to the verifier and repairer instead of the commands.

BraveChicken1 (changes requested):
Please also don't forget the CHANGELOG, README and verifier docs.

src/cli/parameter_checks.rs
Max13245: I'll move it to the cli module then. This would be unpractical, because it would then have to become part of the verifier/register state.

src/cli/parameter_checks.rs
Max13245: Where in the cli module though? Because it's not really part of display and is not a command.

BraveChicken1 (approved):

PR #106, opened 2026-07-02
Add update all and exclude flags

Max13245:
This PR adds two new flags to the `update` command:
- The `--all` flag to update all outdated packages.
- The `--exclude` flag to exclude packages when using the `--all` flag.

The `update` command now also accepts multiple optional packages as input. Note that the new version is now a flag `--new-version` and can only be used when one packages is given.

The `update` command now assumes the latest installed version for the update, unless otherwise specified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Lists all the installed packages.
 Searches a package with `<PACKAGE-NAME>` and shows information based on the package metadata. If the version is given that specific version is searched for.
 
 #### `pit update [<PACKAGE-NAME>[@<VERSION>] ...] [--new-version <NEW-VERSION>] [--all] [--exclude <PACKAGE-NAME> ...]`
-Updates the specified package to the new version, or the latest version if no new version is specified. If multiple packages are specified they are all updated to the latest version (and --new-version cannot be used). If multiple versions of the same package are installed, the `--new-version` flag is required, note that this cannot be done with multiple packages at once. The `--all` flag can be used to update all packages (the latest installed version will be updated). The `--exclude` flag can be used to exclude certain packages when using the `--all` flag.
+Updates the specified package to the new version, or the latest version if no new version is specified. If multiple packages are specified they are all updated to the latest version (and `--new-version` cannot be used). If multiple versions of the same package are installed, the latest installed version is assumed. The `--new-version` flag can be used to specify the new version to install. The `--all` flag can be used to update all packages (the latest installed version will be updated). The `--exclude` flag can be used to exclude certain packages when using the `--all` flag.
 
 #### `pit info <PACKAGE-NAME>[@<VERSION>] [-v] [--tree]`
 Shows info about the specified installed package. If the `-v` option is given, extra information is shown. If the `--tree` option is enabled, the whole dependency tree is shown.
--- a/README.md
+++ b/README.md
@@ -68,8 +68,8 @@ Lists all the installed packages.
 #### `pit search <PACKAGE-NAME>[@<VERSION>]`
 Searches a package with `<PACKAGE-NAME>` and shows information based on the package metadata. If the version is given that specific version is searched for.
 
-#### `pit update <PACKAGE-NAME>[@<VERSION>] [<NEW-VERSION>]`
-Updates the specified package to the new version, or the latest version if no new version is specified. If multiple versions of the same package are installed, the `<VERSION>` option is required.
+#### `pit update [<PACKAGE-NAME>[@<VERSION>] ...] [--new-version <NEW-VERSION>] [--all] [--exclude <PACKAGE-NAME> ...]`
+Updates the specified package to the new version, or the latest version if no new version is specified. If multiple packages are specified they are all updated to the latest version (and --new-version cannot be used). If multiple versions of the same package are installed, the `--new-version` flag is required, note that this cannot be done with multiple packages at once. The `--all` flag can be used to update all packages (the latest installed version will be updated). The `--exclude` flag can be used to exclude certain packages when using the `--all` flag.
 
 #### `pit info <PACKAGE-NAME>[@<VERSION>] [-v] [--tree]`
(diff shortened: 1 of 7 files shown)


src/cli/commands/update.rs
```diff
@@ -20,14 +21,26 @@ use crate::{
...
+/// If multiple packages are specified they are all updated to the latest version.
+/// If multiple versions of the same package are installed, the <VERSION> option is required, note that this
+/// cannot be done with multiple packages at once.
 #[derive(Args, Debug)]
 pub struct UpdateArgs {
-    /// The name of the package to update, with an optional version specified with NAME@VERSION
-    optional_id: OptionalPackageId,
+    /// The packages to update specified with <PACKAGE-NAME>[@<VERSION>] ...
+    #[arg(conflicts_with = "all")]
+    optional_ids: Vec<OptionalPackageId>,
```
```suggestion
    packages: Vec<OptionalPackageId>,
```

src/cli/commands/update.rs
```diff
@@ -20,14 +21,26 @@ use crate::{
...
 
-    /// The version to update to. This can only be a higher version than the current version
+    /// The version to update to. This can only be a higher version than the current version and can only be used when a single package is specified
+    #[arg(long, requires = "optional_ids")]
     new_version: Option<Version>,
+
+    /// Updates all the installed packages to the latest version possible
+    #[arg(long, default_value = "false", conflicts_with_all = ["optional_ids"])]
```
BraveChicken1: Couldn't this just be `conflicts_with`? It seems that `conflicts_with_all` is used to specify multiple conflicting fields.

src/cli/commands/update.rs
```diff
@@ -37,35 +50,111 @@ impl HandleCommand for UpdateArgs {
...
+                error!(msg: "No packages specified to update");
+                exit(1);
+            },
+            false => &self.optional_ids,
+        };
+
+        if optional_ids.is_empty() {
+            error!(msg: "No packages were specified or found to be updatable");
+            exit(1);
         }
```
BraveChicken1: Isn't one of these checks redundant? If you want to keep the first check, the message of the second check could be changed a bit I think.

src/cli/commands/update.rs
```diff
@@ -37,35 +50,111 @@ impl HandleCommand for UpdateArgs {
...
-            None => package_meta.get_latest_version(&Target::current()).unwrap_or_exit(1),
-        };
+        // Check for duplicates, because updating twice will result in an error
+        let duplicates = parameter_checks::get_duplicates(optional_ids);
+        if !duplicates.is_empty() {
+            let mut duplicate_string = String::new();
+            for duplicate in duplicates {
+                duplicate_string.push_str(&duplicate.to_string());
+                duplicate_string.push(' ');
+            }
```
BraveChicken1, line 78:
```suggestion
            let duplicate_string = duplicates.iter().map(|d| d.to_string()).join(" ");
```
But i'm not fully sure

src/installer/installer.rs
```diff
@@ -803,4 +803,27 @@ impl<'a> Installer<'a> {
...
+            // Check if the latest version exists
+            if latest_available_version == latest_installed_version {
+                continue;
+            }
+
+            updatables.push(PackageId::new(package_name.clone(), latest_installed_version.clone()));
+        }
+
+        Ok(updatables)
+    }
```
BraveChicken1, line 838: I'm not sure about the placement of this function here, it does not take the Installer (self) and doesn't really interface with the install functionality either.

README.md
```diff
@@ -68,8 +68,8 @@ Lists all the installed packages.
 #### `pit search <PACKAGE-NAME>[@<VERSION>]`
 Searches a package with `<PACKAGE-NAME>` and shows information based on the package metadata. If the version is given that specific version is searched for.
 
-#### `pit update <PACKAGE-NAME>[@<VERSION>] [<NEW-VERSION>]`
-Updates the specified package to the new version, or the latest version if no new version is specified. If multiple versions of the same package are installed, the `<VERSION>` option is required.
+#### `pit update [<PACKAGE-NAME>[@<VERSION>] ...] [--new-version <NEW-VERSION>] [--all] [--exclude <PACKAGE-NAME> ...]`
+Updates the specified package to the new version, or the latest version if no new version is specified. If multiple packages are specified they are all updated to the latest version (and --new-version cannot be used). If multiple versions of the same package are installed, the `--new-version` flag is required, note that this cannot be done with multiple packages at once. The `--all` flag can be used to update all packages (the latest installed version will be updated). The `--exclude` flag can be used to exclude certain packages when using the `--all` flag.
```
BraveChicken1: ``If multiple versions of the same package are installed, the `--new-version` flag is required``
Why is this required?

BraveChicken1 (changes requested):

src/cli/commands/update.rs
Max13245: Oops, yes, I forgot to delete the second one.

README.md
Max13245: Otherwise you don't know which installed version to update, however since we now assume the latest version for `--all` we could also do that here. Although then I do think we should give the user some kind of sign/warning that multiple versions are installed before updating.

src/cli/commands/update.rs
Max13245: Then it would have to become something like this:
let duplicate_string = duplicates.iter().map(|d| d.to_string()).collect::<Vec<_>>().join(" ");

because the join doesn't work on the iterator.

src/installer/installer.rs
Max13245: Yes I wasn't sure about this either. However, I also didn't know where else to put it. 

The main reason why it doesn't get self is because the installer required a `&mut` to the register instead of just a reference. I could see if a mutable reference works as well, but that seems a bit ugly since it doesn't need it.

BraveChicken1 (approved):

PR #107, opened 2026-07-03
Verifier test script

Max13245:
This PR adds the metadata test scripts as checks in the verifier. It also has a reinstall fix command which is also used by the altered package issue.

### DIFF
--- a/src/integrity/issue.rs
+++ b/src/integrity/issue.rs
@@ -56,6 +56,9 @@ pub enum Issue {
 
     /// A list of packages which have missing symlinks.
     MissingLinks(Vec<PackageName>),
+
+    /// A list of packages for which the test from the repository metadata failed.
+    FailedTest(Vec<PackageId>),
 }
 
 impl Display for Issue {
@@ -190,6 +193,15 @@ impl Display for Issue {
                     writeln!(f, "  - {}", directory.display())?;
                 }
             },
+            Issue::FailedTest(packages) => {
+                writeln!(f, "Failed test")?;
+                let issue_explanation = "The tests for the following packages failed:";
+                writeln!(f, "{issue_explanation}")?;
+
+                for package in packages {
+                    writeln!(f, "  - {package}")?;
+                }
+            },
         }
 
         Ok(())
@@ -222,6 +234,7 @@ impl Issue {
             Issue::InconsistentRegister(_) => {
                 "To fix this issue we try to reconstruct the Register.toml with data still in the Packit directory."
             },
+            Issue::FailedTest(_) => "To fix this issue we try to re-install the package.",
             Issue::AlteredPackage(_) | Issue::MissingPackitGroup => "There is no automatic fix for this issue available yet.",
         }
     }
--- a/src/integrity/issue.rs
+++ b/src/integrity/issue.rs
@@ -234,8 +234,8 @@ impl Issue {
             Issue::InconsistentRegister(_) => {
                 "To fix this issue we try to reconstruct the Register.toml with data still in the Packit directory."
             },
-            Issue::FailedTest(_) => "To fix this issue we try to re-install the package.",
-            Issue::AlteredPackage(_) | Issue::MissingPackitGroup => "There is no automatic fix for this issue available yet.",
+            Issue::FailedTest(_) | Issue::AlteredPackage(_) => "To fix this issue we try to re-install the package.",
+            Issue::MissingPackitGroup => "There is no automatic fix for this issue available yet.",
         }
     }
 }
--- a/src/integrity/issue.rs
+++ b/src/integrity/issue.rs
@@ -199,7 +199,7 @@ impl Display for Issue {
                 writeln!(f, "{issue_explanation}")?;
 
                 for package in packages {
-                    writeln!(f, "  - {package}")?;
+                    writeln!(f, "  - {}", package.style())?;
                 }
             },
         }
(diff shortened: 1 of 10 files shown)


src/integrity/repairer/package.rs
```diff
@@ -324,3 +324,59 @@ pub fn fix_missing_links(missing: Vec<PackageName>, register: &mut PackageRegist
...
+fn reinstall_package(package_id: PackageId, register: &mut PackageRegister, manager: &RepositoryManager, config: &Config) -> Result<()> {
+    let active_directory = config.prefix_directory.join("active");
+    let bin_directory = config.prefix_directory.join("bin");
+    let package_directory = config.prefix_directory.join("packages");
+
+    // Figure out the active version
+    let path = active_directory.join(&package_id.name);
+    let active_target = fs::read_link(&path).err_with_path("read link", path)?;
+    let target_name = active_target.file_name().ok_or(VerifierError::InvalidSymlink)?;
+    let version = Version::from_str(target_name.to_str().ok_or(VerifierError::InvalidUnicodeError)?)?;
```
BraveChicken1:
```suggestion
    let active_version_str = active_target.file_name().ok_or(VerifierError::InvalidSymlink)?;
    let active_version = Version::from_str(active_version_str.to_str().ok_or(VerifierError::InvalidUnicodeError)?)?;
```

src/integrity/repairer/package.rs
```diff
@@ -324,3 +324,59 @@ pub fn fix_missing_links(missing: Vec<PackageName>, register: &mut PackageRegist
...
+
+    // Check if the package should be the active package when installed
+    let active = package_id.version == version;
+
+    // Figure out if symlinked
+    let symlinked = fs::symlink_metadata(bin_directory.join(&package_id.name)).is_ok();
```
BraveChicken1: This check wouldn't always work, since it requires the package to install a binary with the name of the package. For example zlib-ng-compat would fail already. Why don't we use the register?

src/integrity/repairer/package.rs
```diff
@@ -324,3 +324,59 @@ pub fn fix_missing_links(missing: Vec<PackageName>, register: &mut PackageRegist
...
+
+    // Remove the package from the register
+    register.remove_package_version(&package_id);
+
+    // Re-install the package
+    let installer_options = InstallerOptions::default().skip_symlinking(!symlinked).skip_active(!active);
+    let mut installer = Installer::new(config, register, manager, installer_options);
+
+    installer.install(&OptionalPackageId::from(package_id))?;
+
```
BraveChicken1: Shouldn't we insert the original dependents back into the register after installation?

src/integrity/verifier/package.rs
```diff
@@ -560,6 +561,82 @@ fn check_package_dependency_tree(package_id: &PackageId, register: &PackageRegis
...
+    };
+
+    // Get the test script data
+    let version_meta = provider.read_package_version(&package_id.name, &package_id.version)?;
+    let target_bounds = version_meta.get_best_target(&Target::current())?;
+    let target = version_meta.get_target(&target_bounds)?;
+    let script_path = version_meta.get_test_script_path(&target_bounds)?;
+    let script_text = provider.read_file(&package_id.name, &script_path)?;
+    let script_args = version_meta.get_script_args(&target_bounds)?;
+    let Some(script_file) = script_text else { return Ok(false) };
```
BraveChicken1:
```suggestion
    let Some(script_text) = provider.read_file(&package_id.name, &script_path)? else { return Ok(false) };
    let script_args = version_meta.get_script_args(&target_bounds)?;
```

src/integrity/verifier/package.rs
```diff
@@ -560,6 +561,82 @@ fn check_package_dependency_tree(package_id: &PackageId, register: &PackageRegis
...
+                return Ok(false);
+            },
+        }
+    }
+
+    // Run the test script and only return true if the test itself failed
+    match scripts::run_test_script(&script_data, &read_files) {
+        Ok(_) => Ok(false),
+        Err(ScriptError::ScriptFailed(..)) => Ok(true),
+        Err(e) => return Err(crate::integrity::error::VerifierError::ScriptError(e)),
```
BraveChicken1:
```suggestion
        Err(e) => return Err(VerifierError::ScriptError(e)),
```

src/integrity/issue.rs
```diff
@@ -190,6 +193,15 @@ impl Display for Issue {
                     writeln!(f, "  - {}", directory.display())?;
                 }
             },
+            Issue::FailedTest(packages) => {
+                writeln!(f, "Failed test")?;
+                let issue_explanation = "The tests for the following packages failed:";
+                writeln!(f, "{issue_explanation}")?;
+
+                for package in packages {
+                    writeln!(f, "  - {package}")?;
```
BraveChicken1:
```suggestion
                    writeln!(f, "  - {}", package.style())?;
```

BraveChicken1 (changes requested):

src/integrity/repairer/package.rs
Max13245: This implementation is partially taken from the original fix for the broken register. So then we couldn't do this, now we can. I'll change it.

src/integrity/repairer/package.rs
Max13245: No, because they are never deleted, that's the reason for the manual uninstall.
The other option would be to uninstall and the re-insert the dependents, but that's also kind of buggy.

BraveChicken1 (approved):

PR #108, opened 2026-07-04
Extend metadata fields

BraveChicken1:
This PR extends the available fields in the metadata:
- Adds the `required_packit_version` fields in `repository.toml`, `package.toml` and `targets.toml` to specify the minimum required Packit version for installing the package.
- Adds the `conflicts_with` field in package.toml` to describe conflicting packages. Two conflicting packages cannot be symlinked at the same time.
- Adds the `size` field to sources, to specify the size of the source in bytes.
- Adds deprecation information fields, to allow specifying deprecation and disabling dates of packages or specific versions.
- Changes the metadata resolving algorithms, showing a clear reason when a package cannot be found and trying to find the best suitable package and version.


src/repositories/manager.rs
```diff
@@ -248,9 +348,110 @@ impl<'a> RepositoryManager<'a> {
...
+    /// Gets the latest supported package version for the given package metadata.
+    pub fn read_latest_supported_version(&self, repository_id: &str, package: &PackageMeta, target: &Target) -> Result<PackageVersionMeta> {
+        let supported_versions = package.get_supported_versions(target)?;
+        let supported_versions = supported_versions.into_iter().rev();
+
+        self.read_latest_supported_version_impl(repository_id, package, supported_versions)
+    }
+
+    /// Gets the latest supported package version for the given package metadata that satisfies the given dependency.
+    // TODO: change name of this function
```
Max13245: You left a TODO here.

src/repositories/manager.rs
```diff
@@ -128,31 +139,78 @@ impl<'a> RepositoryManager<'a> {
...
 
-            // Check if package contains the target
-            if package.get_latest_version(&Target::current()).is_err() {
+            // Check package compatibility
+            if let Some(reason) = self.check_package_compatibility(&package) {
+                not_found_reasons.insert(repository_id, reason);
                 continue;
             }
 
+            // TODO: do we want priority for packages that are not deprecated?
```
Max13245: You left a TODO here. I think this one has already been decided and solved?

src/repositories/manager.rs
```suggestion
    /// Reads package and package version metadata of the given package. When only a package name is given the latest supported version is used.
```

src/repositories/types/repository.rs
```diff
@@ -1,14 +1,15 @@
...
+use crate::{installer::types::Version, repositories::types::Licenses};
 
 /// Represents the repository metadata, containing repository information.
-#[derive(Deserialize, Serialize, Debug)]
+#[derive(Deserialize, Serialize, Clone, Debug)]
 pub struct RepositoryMeta {
     pub name: String,
     pub description: String,
     pub maintainers: Vec<String>,
+    pub required_packit_version: Option<Version>,
```
Max13245: I think that the minimum required Packit version should not be optional on the repository metadata. Then there is always a minimal Packit version.

src/repositories/error.rs
```diff
@@ -11,10 +15,16 @@ pub enum RepositoryError {
         repository_id: String,
     },
 
-    #[error("Cannot find package '{package_name}' with version '{}' in any repository", version.as_deref().unwrap_or("any"))]
+    #[error("Repository '{repository_id}' not supported")]
```
Max13245:
```suggestion
    #[error("Repository '{repository_id}' not supported. Packit version '{}' was needed, but only a lower version '{}' was used.")]
```

src/cli/commands/link.rs
Max13245, line 80: Maybe we could make these checks warnings when `--force` is specified, combined with the question if the user is sure that they want to do a force link. 

I'm not sure if that's something for this PR though.

src/repositories/manager.rs
```diff
@@ -248,9 +348,110 @@ impl<'a> RepositoryManager<'a> {
...
+        Err(RepositoryError::PackageNotFoundError {
+            package_name: package.name.to_string(),
+            version: None,
+            reason: PackageNotFoundReason::get_primary_reason(reasons.iter()),
+        })
+    }
+
+    /// Iterates over the repositories rank. Filters the unsupported repositories from the list.
+    /// Returns an iterator that only contains the ids of working repositories.
+    pub fn iter_filtered_repositories_rank(&self) -> impl Iterator<Item = &String> {
```
```suggestion
    /// Iterates over the repositories rank. Filters the unsupported repositories from the list.
    /// Returns an iterator that only contains the ids of supported repositories.
    pub fn iter_supported_repositories_rank(&self) -> impl Iterator<Item = &String> {
```

src/cli/commands/search.rs
```diff
@@ -134,5 +133,31 @@ impl SearchArgs {
...
+                Some(reason) => format!(" with reason '{reason}'"),
+                None => String::default(),
+            };
+
+            match deprecation.deprecated_from <= Date::now() {
+                true => warning!("This package version is deprecated since {}{reason}", deprecation.deprecated_from),
+                false => warning!("This package version will be deprecated at {}{reason}", deprecation.deprecated_from),
+            }
+        }
     }
```
Max13245: Most of our warnings are at the top of the displayed text I think. Maybe move this to after the retrieval of the latest version.

Also this logic is found in several places, maybe make this a separate display util. Then we also ensure consistent prints.

src/cli/display/not_found.rs
```diff
@@ -79,14 +81,27 @@ pub fn register_package_version(package_id: &PackageId, register: &PackageRegist
 }
 
 /// Shows an error that the package version cannot be found (in the repository) and a fuzzy alternative. Then exits at the end.
```
Max13245:
```suggestion
/// Shows an error that the package version cannot be found (in the repository) and a fuzzy alternative when the given reason is `PackageNotFoundReason::NotFound`. Then exits at the end.
```

src/cli/display/not_found.rs
```diff
@@ -54,12 +53,15 @@ pub fn register_package(package_name: &PackageName, register: &PackageRegister)
 }
 
 /// Shows an error that the package (name) cannot be found (in the repository) and a fuzzy alternative. Then exits at the end.
```
Max13245:
```suggestion
/// Shows an error that the package (name) cannot be found (in the repository) and a fuzzy alternative when the given reason is `PackageNotFoundReason::NotFound`. Then exits at the end.
```

src/cli/display/not_found.rs
```diff
@@ -79,14 +81,27 @@ pub fn register_package_version(package_id: &PackageId, register: &PackageRegist
...
+        let fuzzy_match = fuzzy::repository_search(manager, &package_id.name).unwrap_or_exit(1);
+        if let Some(fuzzy_match) = fuzzy_match {
+            println!("Did you mean: '{fuzzy_match}'?");
+        }
     }
 
     exit(1);
 }
+
+/// Shows an error that the package (with optional version) cannot be found and a fuzzy alternative. Then exits at the end.
```
Max13245:
```suggestion
/// Shows an error that the package (with optional version) cannot be found and a fuzzy alternative when the given reason is `PackageNotFoundReason::NotFound`. Then exits at the end.
```

src/cli/display/not_found.rs
```diff
@@ -79,14 +81,27 @@ pub fn register_package_version(package_id: &PackageId, register: &PackageRegist
...
+/// When the package_id contains a version, a version suggestion is also shown.
```
Max13245:
```suggestion
/// When the package_id contains a version, a version suggestion is shown when the package name is correct.
```

src/repositories/types/package.rs
```diff
@@ -33,47 +39,22 @@ impl PackageMeta {
         let supported = self.supported_versions.get(target).ok_or(RepositoryError::TargetError)?;
 
         // The versions vec isn't necessary in order, so we need to keep track of the current highest version
```
Max13245:
```suggestion
        // Get the supported versions
```

src/repositories/types/package.rs
```diff
@@ -19,12 +19,18 @@ pub struct PackageMeta {
...
     pub supported_versions: HashMap<TargetBounds, VersionIntervals>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub deprecation: Option<DeprecationInfo>,
 }
 
 impl PackageMeta {
-    /// Gets the latest supported version for the current target.
-    pub fn get_latest_version(&self, target: &Target) -> Result<&Version> {
+    pub fn get_supported_versions(&self, target: &Target) -> Result<Vec<&Version>> {
```
Max13245, line 34: Please add documentation for this function.

src/repositories/error.rs
```diff
@@ -55,3 +68,66 @@ pub enum RepositoryError {
...
+                // If the primary is Disabled, only use the new one if the next if UnsupportedTarget or NotSupported, or a newer disable
+                PackageNotFoundReason::Disabled { since, .. } => {
+                    match next {
+                        PackageNotFoundReason::UnsupportedTarget | PackageNotFoundReason::NotSupported { .. } => primary = next.clone(),
+                        PackageNotFoundReason::Disabled { since: since_next, .. } if since_next > since => primary = next.clone(),
+                        _ => {},
+                    };
+                },
+
+                // If the current primary is not found, always use the new one
```
Max13245:
```suggestion
                // If the current primary is `NotFound`, always use the new one
```

src/repositories/error.rs
```diff
@@ -55,3 +68,66 @@ pub enum RepositoryError {
...
+
+impl PackageNotFoundReason {
+    /// Gets the primary reason from an iterator over reasons
+    /// If the iterator has no items, PackageNotFoundReason::NotFound is returned.
+    pub fn get_primary_reason<'a>(reasons: impl Iterator<Item = &'a PackageNotFoundReason>) -> PackageNotFoundReason {
+        let mut primary = PackageNotFoundReason::NotFound;
+
+        for next in reasons {
+            match &primary {
+                // If the primary is UnsupportedTarget, only use new one if the next is not NotFound or Disabled
```
Max13245:
```suggestion
                // If the primary is UnsupportedTarget, only use the new one if the next is not `NotFound` or `Disabled`
```

src/repositories/error.rs
```diff
@@ -55,3 +68,66 @@ pub enum RepositoryError {
...
+                PackageNotFoundReason::UnsupportedTarget => {
+                    if !matches!(next, PackageNotFoundReason::NotFound | PackageNotFoundReason::Disabled { .. }) {
+                        primary = next.clone();
+                    }
+                },
+
+                // If the primary is NotSupported, only use new one if next is NotSupported and the required version is lower
```
Max13245:
```suggestion
                // If the primary is NotSupported, only use the new one if next is `NotSupported` and the required version is lower
```

src/repositories/error.rs
```diff
@@ -55,3 +68,66 @@ pub enum RepositoryError {
...
+                PackageNotFoundReason::NotSupported { requires } => {
```
Max13245, line 109: Shouldn't `NotSupported` also set the primary when:
```
if !matches!(next, PackageNotFoundReason::NotFound | PackageNotFoundReason::Disabled { .. }) {
    primary = next.clone();
}
```

src/repositories/manager.rs
```diff
@@ -38,29 +41,39 @@ impl<'a> RepositoryManager<'a> {
                 continue;
             };
 
+            // Read repository metadata, or skip repository is metadata cannot be read
```
```suggestion
            // Read repository metadata, or skip repository if the metadata cannot be read
```

src/repositories/manager.rs
```diff
@@ -128,31 +139,78 @@ impl<'a> RepositoryManager<'a> {
...
 
         Err(RepositoryError::PackageNotFoundError {
             package_name: package.to_string(),
             version: None,
+            reason: PackageNotFoundReason::get_primary_reason(not_found_reasons.values()),
         })
     }
 
     /// Reads package metadata of the given package from the given repository, containing information about the package.
     /// Does not check if the data contains the current target.
```
```suggestion
    /// Also checks for package compatibility.
```

src/repositories/manager.rs
```diff
@@ -128,31 +139,78 @@ impl<'a> RepositoryManager<'a> {
...
+                package_name: package.to_string(),
+                version: None,
+                reason,
+            });
+        }
+
+        Ok(package_meta)
+    }
+
+    /// Checks if a package is compatible.
```
```suggestion
    /// Checks if a package is compatible, a package is compatible if:
    /// - It's not deprecated
    /// - It contains the current target
    /// - The package is supported by the current Packit version
```

src/repositories/manager.rs
```diff
@@ -169,14 +227,15 @@ impl<'a> RepositoryManager<'a> {
...
+            // Check package version compatibility
+            if let Some(reason) = self.check_package_version_compatibility(&package, target) {
+                not_found_reasons.insert(repository_id, reason);
                 continue;
             }
 
             // Validate the package before returning
             if package.has_conflicts() {
-                return Err(RepositoryError::ValidationError("Package has dependency conflicts.".to_string()));
+                return Err(RepositoryError::ValidationError("Package has conflicts in metadata.".to_string()));
```
Max13245, line 239:
```suggestion
                return Err(RepositoryError::ValidationError("Package has conflicts according to the metadata.".to_string()));
```

Otherwise it sounds like the metadata itself has a conflict.

src/repositories/manager.rs
```suggestion
    /// Also checks for package version compatibility.
```

src/repositories/manager.rs
```diff
@@ -193,16 +253,56 @@ impl<'a> RepositoryManager<'a> {
...
         // Validate the package before returning
-        let package = provider.read_package_version(&package_id.name, &package_id.version)?;
         if package.has_conflicts() {
             return Err(RepositoryError::ValidationError("Package has dependency conflicts.".to_string()));
         }
 
         Ok(package)
     }
 
+    /// Checks if a package version is compatible.
```
```suggestion
    /// Checks if a package version is compatible, a package version is compatible if:
    /// - It's not deprecated
    /// - It contains the current target
    /// - The package version is supported by the current Packit version
```

If the package compatibility check is added:
/// - The package (not package version) is also compatible

src/repositories/manager.rs
```diff
@@ -193,16 +253,56 @@ impl<'a> RepositoryManager<'a> {
...
+    /// Returns a Some(PackageNotFoundReason) if the package version is not compatible, None otherwise.
+    fn check_package_version_compatibility(&self, package_version: &PackageVersionMeta, target: &Target) -> Option<PackageNotFoundReason> {
```
Max13245, line 281: Shouldn't we check here if the package (not package version) is also compatible? If that's not compatible then this shouldn't be either right?

src/repositories/manager.rs
```diff
@@ -248,9 +348,110 @@ impl<'a> RepositoryManager<'a> {
...
+    fn read_latest_supported_version_impl<'v>(
+        &self,
+        repository_id: &str,
+        package: &PackageMeta,
+        mut supported_versions: impl Iterator<Item = &'v Version>,
+    ) -> Result<PackageVersionMeta> {
+        let mut reasons = Vec::new();
+        let mut latest_deprecated: Option<PackageVersionMeta> = None;
+        loop {
+            let Some(latest_version) = supported_versions.next() else { break };
```
```suggestion
        while let Some(latest_version) = supported_versions.next() {
```

src/repositories/manager.rs
```diff
@@ -248,9 +348,110 @@ impl<'a> RepositoryManager<'a> {
...
+            Err(RepositoryError::PackageNotFoundError { reason, .. }) => Err(RepositoryError::DependencyNotFoundError {
+                dependency: dependency.to_string(),
+                reason,
+            }),
+            Err(e) => Err(e),
+        }
+    }
+
+    /// Implementation of the read latest supported version function.
+    /// Resolves the latest supported version based on compatibility checks and implements deprecation resolving.
```
Max13245, line 385:
```suggestion
    /// Implementation of the read latest supported version function.
    /// Resolves the latest supported version based on compatibility checks and implements deprecation resolving.
    /// Note that it assumes a list of supported versions that is in order of low to high.
```

src/repositories/manager.rs
```diff
@@ -248,9 +348,110 @@ impl<'a> RepositoryManager<'a> {
...
+        })
+    }
+
+    /// Iterates over the repositories rank. Filters the unsupported repositories from the list.
+    /// Returns an iterator that only contains the ids of working repositories.
+    pub fn iter_filtered_repositories_rank(&self) -> impl Iterator<Item = &String> {
+        self.config.repositories_rank.iter().filter(|x| !self.unsupported_repositories.contains_key(*x))
+    }
+
+    /// Gets the list of repositories that are not supported by this Packit version.
```
```suggestion
    /// Gets the map of repositories that are not supported by this Packit version.
```

src/repositories/manager.rs
```diff
@@ -248,9 +348,110 @@ impl<'a> RepositoryManager<'a> {
...
+    pub fn get_unsupported_repositories(&self) -> &HashMap<String, RepositoryMeta> {
+        &self.unsupported_repositories
+    }
+
     /// A helper method to het the metadata provider.
```
```suggestion
    /// A helper method to get the metadata provider.
```

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - The update command now accepts multiple packages as input instead of just one.
 - The `--all` flag for the update command flag to update all packages which are not up-to-date.
 - The `--exclude` flag for the update command to exclude certain packages when using the `--all` flag.
+- The `required_packit_version` field to describe the minimum required Packit version for a package in `repository.toml`, `package.toml` and `targets.toml` metadata files.
+- The `conflicts_with` fields to describe conflicting packages in the `package.toml` metadata file. Two conflicting packages cannot be symlinked at the same time.
+- The `size` field in the source fields in the `targets.toml` metadata file.
+- Deprecation information fields, to allow specifying deprecation and disabling dates of packages or specific versions.
```
Max13245:
```suggestion
- Deprecation information fields (deprecated_from, disabled_from and reason), to allow specifying deprecation and disabling dates of packages or specific versions.
```

CHANGELOG.md
```diff
@@ -26,6 +30,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - The `gnubin` directory of a package is now also symlinked into `<prefix>/gnubin`.
 - The `check` and `fix` commands now accept `[<PACKAGE-NAME>[@<VERSION>] ...]` instead of just `[<PACKAGE-NAME>@<VERSION> ...]`.
 - The `update` command now updates the latest install version, unless otherwise specified.
+- The package metadata resolving algorithm now finds matching packages in more cases and shows a clear reason when a package cannot be found.
```
Max13245: It's not really clear to me why our resolving algorithm is now finding packages in more cases?

Max13245 (changes requested):
Really nice, I'm exited about these changes! Here is my essay of review content :)

Maybe it's nice for the user to have a way to ignore deprecation. One of the greatest advantages of Packit is that it's flexible in it's dependencies and versioning. This would make that less so. 

I think that the symlinker itself should check for conflicts before doing symlinks. The `create_symlinks` method could then maybe accept a `InstalledPackageVersion` is input instead. The symlinker is used in a lot of different places, so it's easy to forget to check conflicts.

When adding a new repository with the config command with an incompatible Packit version it might be nice to warn the user that the repository they are trying to add isn't compatible with their current Packit version.

The minimal Packit version change makes it even more important that we store the metadata of installed packages. Otherwise a user might not be able to uninstall a package if they have the wrong Packit version. (Even worse, if they can't update Packit because they have the wrong Packit version)

When we create the provider manually, with the help of the repository manager we now don't check for compatibility. This could be quite a big problem I think? I think we'll need to handle creation of repositories with url's through the manager. This might also clean up some code.

src/repositories/manager.rs
BraveChicken1: I'll probably remove this TODO then, I don't really have a better name for this function

src/repositories/manager.rs
BraveChicken1: Deprecation resolving is implemented for getting the latest package version only, for getting a package it is not implemented. If we would implement it here too, a deprecated package in repository a would mean the package would be fetched from repository b if that is not deprecated.

src/cli/commands/link.rs
BraveChicken1: Let's do that in a separate PR

src/repositories/manager.rs
BraveChicken1: I called it filtered because that name is already future proof for other repository filters, however I will change it to supported now.

src/cli/commands/search.rs
BraveChicken1: The warnings at the top are warnings from other parts of Packit, putting the warnings at the bottom makes these warnings better visible.

I could move it to a util, but the message is slightly different in both cases.

src/repositories/error.rs
BraveChicken1: The order here is NotFound - Disabled - UnsupportedTarget - NotSupported.
So when the primary reason is already NotSupported, we only want to change the primary reason if we find a NotSupported with a lower required version.

src/repositories/manager.rs
BraveChicken1: The metadata does have conflicts, the has_conflicts function checks if the metadata is valid.

src/repositories/manager.rs
BraveChicken1: That is checked when the PackageMeta is fetched, fetching the PackageMeta and PackageVersionMeta is done separately. We could still improve this in a future version.

CHANGELOG.md
BraveChicken1: We check for constraints more consistently now, but I'll change this into just `The package metadata resolving algorithm is improved and now shows a clear reason when a package cannot be found.`

BraveChicken1:
If a user really wants to install a deprecated version, the user can just install a specific version of a package. That is still supported and fully working. The current implementation just prevents Packit from choosing a deprecated version as latest version.

For the symlinker, I wanted to this but noticed it complicated the code a bit, I'll take a look at it again.

I'll add the warning to the config repository add command

The storing of the metadata shouldn't really solve a lot of Packit compatibility issues here, although I would be a really good addition. This would also solve the problem with creating providers directly, so I'm not sure if we want to solve this now already, or together with metadata storing.

src/cli/commands/search.rs
Max13245: Alright

Yes, but in case of the search command you could use the package version that was found right?

src/repositories/error.rs
BraveChicken1: I don't have this version information when creating the error, I only know that the repository is in the list of unsupported repositories.

src/repositories/error.rs
Max13245: If the order is NotFound - Disabled - UnsupportedTarget - NotSupported, then why does UnsupportedTarget "go back to" NotFound - Disabled?
(And Disabled to UnsupportedTarget?)

src/repositories/manager.rs
Max13245: Oops, my bad, thought this was the package conflicts thing. Maybe we should rename the function to `has_metadata_conflicts` or `has_meta_conflicts`?

src/cli/commands/search.rs
BraveChicken1: This is used in the search command and in the install command, in the search command we use `This package` and in the install command `Package '<optional_id>'`.

src/repositories/manager.rs
Max13245: Well, I think it's also important to do it here, because then you don't rely on all the times this function is called to implement it right. And it's pretty important to check this, because otherwise you could say that the package is deprecated, but specific versions are still okay to use.

Max13245:
True

Great

Nice

Well, if we store the metadata, then you would be using older metadata which is compatible with your Packit version. Although this actually raises a different question about the implementation of local metadata storage (how is it stored and do we update when updating Packit). This would be a very big change though, so definitely for a different PR and next version.

src/repositories/manager.rs
Max13245: Ahh, like that. Mhmmm

I would say we can definitely leave it as is for now. If a package is deprecated in one repository it's probably also deprecated in the other repository, if not you could also ask yourself; is this repo up-to-date? Also, the first repository that is checked is also the users preferred repo, so if that one says the package is deprecated and another says it's not, the user has already kind of chosen indirectly that they would rather use data from the first repo.

src/repositories/error.rs
Max13245: Mhmm, in that case maybe this, to still make clear that if they update their Packit version it could solve their problem:
#[error("Repository '{repository_id}' not supported, the current Packit version is too low.")]

src/cli/commands/search.rs
Max13245: Yes, I know, but you could change the text in the search command to use the package version that was found with the search command right? And then a general implementation would be possible.

src/repositories/error.rs
BraveChicken1: I could do that, but that would not be future proof for future other repository compatibility constraints.

src/repositories/error.rs
BraveChicken1: They don't go back, that only check that the next is one of the reasons with more priority.

src/repositories/manager.rs
BraveChicken1: Or `is_metadata_valid`?

src/repositories/manager.rs
BraveChicken1: But this function is only called within the RepositoryManager to validate the PackageVersionMeta is compatible. This is then later combined with the PackageMeta compatibility check.

src/repositories/error.rs
Max13245: I would say change it for now, then we can change it when necessary in the future (as it's a small change). Or, if future proving is very important for you, you could also introduce a reason field, like in the other errors.

src/repositories/error.rs
Max13245: What I mean is that with this implementation I think we can go from `Disabled` to `UnsupportedTarget`, but also from `UnsupportedTarget` to `Disabled`.

src/repositories/manager.rs
Max13245: Yes that's better!

src/repositories/manager.rs
Max13245: But it's also used in the `read_repo_package_version` and `read_package_version` functions, which are used outside of the manager right?

src/repositories/manager.rs
BraveChicken1: Then I'll remove the TODO for now.

src/repositories/error.rs
BraveChicken1: From `Disabled` to `UnsupportedTarget` is expected behaviour, the other way around is not possible, as we check if the next does **not** match `NotFound` or `Disabled`.

src/repositories/manager.rs
BraveChicken1: I changed it. A good future improvement would probably be to have better error logging here, showing also what the conflicting thing in the metadata is.

src/repositories/manager.rs
BraveChicken1: Yes, they are used outside of the manager, but only to get a package version. It depends on the calling case if we need to fetch the PackageMeta too. But the read_package functions already do the compatibility checks for PackageMeta.

src/repositories/error.rs
Max13245: Ahh I see it now, nvm then :)

src/repositories/manager.rs
Max13245: Yes, but if a package is deprecated, then all package versions are also deprecated right? So then when only using `read_package_version` we still need to check if the package is not deprecated.

BraveChicken1:
All requested changes are now implemented. For a future PR, we could look into improving the resolving of packages even more. We should probably look better into the relation between PackageMeta and PackageVersionMeta.

Max13245 (approved):

PR #111, opened 2026-07-06
Improved Packit install scripts

Max13245:
This PR adds:
- Better dialogue and information printing
- Separate ask methods in both Unix and Windows scripts (which are case insensitive)
- Cleanup functions in case of errors during the Packit installation
- A check for internet connectivity after the prebuild download fails

Note that the Windows install script cannot execute the cleanup function when a ctrl+C interrupt happens, as this is at the very least very cumbersome... (powershell is probably needed).

### DIFF
--- a/install.bat
+++ b/install.bat
@@ -25,17 +25,17 @@ set "CONFIG_DIR=C:\Program Files\packit"
 echo Prefix directory: %PREFIX_DIR%
 echo Config directory: %CONFIG_DIR%
 
+REM TODO: Use choice command everywhere
+REM Ask the user for administrator rights
+choice /M "The Packit install script requires root to modify '%PREFIX_DIR%' and '%CONFIG_DIR%', do you wish to continue? (Y/n) "
+if ERRORLEVEL 2 (
+    echo Packit install cancelled
+    goto cleanup
+)
+
 REM Check for administrator privileges 
 fltmc >nul 2>&1
 if ERRORLEVEL 1 (
-    echo The Packit install requires administrator privileges to modify '%PREFIX_DIR%' and '%CONFIG_DIR%'.
-    choice /M "Do you wish to continue as administrator?"
-
-    if ERRORLEVEL 2 (
-        echo Packit install cancelled
-        goto cleanup
-    )
-
     REM Rerun the script with elevated permissions (this will first prompt the user)
     powershell -Command "Start-Process cmd -Verb RunAs -ArgumentList '/k \"%~f0\"'"
 
--- a/install.bat
+++ b/install.bat
@@ -31,7 +31,7 @@ goto cleanup_end
 set STATUS_CODE=%ERRORLEVEL%
 
 popd
-echo Remove installed Packit files
+echo Removing installed Packit files
 
 REM Remove the prefix directory if `PREFIX_DIR` exists
 if exist "%PREFIX_DIR%" (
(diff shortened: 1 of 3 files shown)


install.bat
```diff
@@ -1,170 +1,256 @@
...
+    goto cleanup
+)
+
+REM Check for administrator privileges 
+fltmc >nul 2>&1
+if ERRORLEVEL 1 (
+    REM Rerun the script with elevated permissions (this will first prompt the user)
+    powershell -Command "Start-Process cmd -Verb RunAs -ArgumentList '/k \"%~f0\"'"
+
+    exit /b
```
BraveChicken1, line 88: Shouldn't there be an exit code here?

install.bat
```diff
@@ -1,170 +1,256 @@
...
             if defined PROCESSOR_ARCHITEW6432 (
+                echo Installing cargo from 'https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-msvc/rustup-init.exe'
                 curl --proto "=https" --tlsv1.2 -sSfL https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-msvc/rustup-init.exe --output rustup-init.exe
+                if ERRORLEVEL 1 goto cleanup
             ) else (
+                echo Installing cargo from 'https://static.rust-lang.org/rustup/dist/i686-pc-windows-msvc/rustup-init.exe'
                 curl --proto "=https" --tlsv1.2 -sSfL https://static.rust-lang.org/rustup/dist/i686-pc-windows-msvc/rustup-init.exe --output rustup-init.exe
+                if ERRORLEVEL 1 goto cleanup
             )
         )
```
BraveChicken1: Wouldn't it be easier to set a variable, and do a single echo and curl after?

install.sh
```diff
@@ -12,11 +12,64 @@ get_answer() {
...
+    else
+        printf "$QUESTION? (y/N) "
+        answer=$(to_lower "$(get_answer)")
+        ! { [ "$answer" = "y" ] || [ "$answer" = "yes" ]; }
+    fi
+}
+
+# Removes all created files in case of an error.
+cleanup() {
+    # Don't cleanup of `SHOULD_CLEANUP` is not true
```
BraveChicken1:
```suggestion
    # Don't cleanup if `SHOULD_CLEANUP` is not true
```

install.sh
```diff
@@ -12,11 +12,64 @@ get_answer() {
...
+    if [ ${SHOULD_CLEANUP-} -eq 0 ]; then
+        exit 0
+    fi
+
+    echo "Remove installed Packit files"
```
BraveChicken1:
```suggestion
    echo "Removing installed Packit files"
```

install.bat
```diff
@@ -1,170 +1,256 @@
...
+
+:ask_end
+
+REM Removes all created files in case of an error and exits with the appropriate status code.
+goto cleanup_end
+:cleanup
+set STATUS_CODE=%ERRORLEVEL%
+
+popd
+echo Remove installed Packit files
```
BraveChicken1:
```suggestion
echo Removing installed Packit files
```

install.sh
```diff
@@ -58,6 +113,26 @@ else
...
+# Execute sudo without doing anything
+sudo true
+
+# Exit early with code 0 if there already is a version of Packit installed
+# Note that we can't rely on the `packit init` command, because we don't know if it fails because of an already existing config file
+if [ -f "$CONFIG_DIR/Config.toml" ]; then
+    echo "Packit already seems to be installed, config file found in '$CONFIG_DIR'"
+    SHOULD_CLEANUP=0
+    exit 0
+fi
```
BraveChicken1, line 135: Shouldn't we also check for the prefix or register? The init command also checks multiple things.

BraveChicken1 (changes requested):
Looks really good!

install.bat
Max13245: It's used to return while preserving the ERRORLEVEL from the created sub-process.

BraveChicken1 (approved):

PR #112, opened 2026-07-07
Add extra script outputs

BraveChicken1:
This PR adds support for additional script outputs:
- file descriptor 3 can be used to output verbose output, Packit will control the destination of it.
- when the build script exits with an error, the last 10 lines are now shown.


src/installer/scripts.rs
```diff
@@ -250,3 +300,47 @@ pub const SCRIPT_EXTENSION: &str = "sh";
 
 #[cfg(target_os = "windows")]
 pub const SCRIPT_EXTENSION: &str = "bat";
+
+#[cfg(any(target_os = "macos", target_os = "linux"))]
+fn bind_extra_outputs(command: &mut Command, script_data: &ScriptData) -> Result<()> {
```
Max13245, line 307: Please make some function comments (Also for the Windows equivelant).

src/installer/scripts.rs
```diff
@@ -250,3 +300,47 @@ pub const SCRIPT_EXTENSION: &str = "sh";
...
+        true => libc::STDOUT_FILENO,
+        false => fs::File::options().write(true).open("/dev/null").err_operation("open /dev/null")?.into_raw_fd(),
+    };
+
+    // SAFETY: only dup2 and close operations are called in the subprocess which are async-signal-safe.
+    // SAFETY: ownership of the /dev/null file descriptor is moved into the subprocess and closed there.
+    // SAFETY: errors of dup2 are propagated back to the spawn function.
+    unsafe {
+        command.pre_exec(move || {
+            if libc::dup2(source_verbose_fd, 3) == -1 {
```
Max13245, line 320: Do we want an extra safety here which checks if fd 3 is already open (as it will close silently, with errors and is in an `unsafe` block)?

docs/metadata.md
```diff
@@ -197,3 +197,7 @@ Please note that the build script output is only shown to the user when the verb
 The script arguments that are defined in the metadata are passed to the script as environment variable as `PACKIT_ARGS_<argument-name>`.
 
 The environment of build scripts are managed more extensively to ensure reproducible builds.
+
+Scripts have the ability to use file descriptor 3 to print verbose output, this output is only shown to user when verbose mode is turned on. Scripts should only print absolute necessary output to stdout and stderr, other output should be redirected to this verbose stream.
```
Max13245:
```suggestion
Scripts have the ability to use file descriptor 3 to print verbose output, this output is only shown to the user when verbose mode is turned on. Scripts should only print absolutely necessary output to stdout and stderr, other output should be redirected to this verbose stream.
```

Max13245 (changes requested):

src/installer/scripts.rs
BraveChicken1: It shouldn't really matter since fd 3 is not opened anywhere else and the only thing that will be run is the script.

Max13245: True, it shouldn't, I'm just worried that if, for some reason, in the future if this changes this might be overlooked. Maybe at least a comment?

BraveChicken1: My reasoning here is that we always want to have this stream available at fd 3, so it doesn't really matter if there was another stream here. I can add a check and a warning just to be sure

Max13245: Thnx :)

Max13245 (approved):

PR #113, opened 2026-07-09
Improve Packit UI

Max13245:
This PR implements improvements for the Packit UI.
- Aligned printing
- Standard text colloring
- Nice list printing
- Better spinner message handeling


src/builder/builder.rs
```diff
@@ -206,19 +207,16 @@ impl<'a> Builder<'a> {
...
-            "Downloading 'patch {id}' of '{package_id}' from repository '{repository_id}' successful"
-        ));
+        spinner.finish();
 
         Ok(file)
     }
 
     /// Downloads a file from the url, or one of the mirrors. Checks against a checksum and shows a spinner.
-    fn download_file(&self, url: &str, mirrors: &[String], checksum: &Checksum, download_description: &str) -> Result<Bytes> {
+    fn download_file(&self, url: &str, mirrors: &[String], checksum: &Checksum, download_description: String) -> Result<Bytes> {
```
BraveChicken1:
```suggestion
    fn download_file(&self, url: &str, mirrors: &[String], checksum: &Checksum, download_description: &str) -> Result<Bytes> {
```

Please don't use a string here, since we don't need the allocated string in this function.

src/cli/commands/config.rs
```diff
@@ -162,12 +167,15 @@ impl ConfigArgs {
...
-                println!("Maintainers: {}", metadata.maintainers.join(", "));
-                println!("Repository provider: {}, url: {}", repository.provider, repository.url);
-                println!("Required Packit Version: {}", metadata.required_packit_version.to_string().red());
+                println!("{} ({repository_id}) {}", metadata.name.bold().blue(), "NOT SUPPORTED".bold().red());
+                println!("{}", metadata.description.italic().cyan());
+                let mut pair_aligner = PairAligner::new();
+                pair_aligner.add("License", &metadata.license);
+                pair_aligner.add("Maintainers", metadata.maintainers.join(", "));
+                pair_aligner.add("Repository provider", &repository.provider);
+                pair_aligner.add("Repository url", &repository.url);
```
BraveChicken1, line 176: Maybe it would be nicer to format it as:
Repository url: {url} ({provider})

src/cli/commands/config.rs
```diff
@@ -184,11 +192,14 @@ impl ConfigArgs {
...
-            println!("License: {}", metadata.license);
-            println!("Maintainers: {}", metadata.maintainers.join(", "));
-            println!("Repository provider: {}, url: {}", repository.provider, repository.url);
-            println!("Required Packit Version: {}", metadata.required_packit_version.to_string().green());
+            println!("{}", metadata.description.italic().cyan());
+            let mut pair_aligner = PairAligner::new();
+            pair_aligner.add("License", metadata.license);
+            pair_aligner.add("Maintainers", metadata.maintainers.join(", "));
+            pair_aligner.add("Repository provider", &repository.provider);
+            pair_aligner.add("Repository url", &repository.url);
```
BraveChicken1, line 200: Same comment as above.

src/cli/commands/info.rs
```diff
@@ -74,20 +81,18 @@ impl HandleCommand for InfoArgs {
...
-        println!("Active version: {}", package.active_version);
-        println!("Symlinked: {}", package.symlinked);
+        println!("{}", self.package.name.style());
+        println!("{}", package.description.italic().cyan());
+
+        let mut pair_aligner = PairAligner::new();
+        pair_aligner.add("Homepage", package.homepage.display());
+        pair_aligner.add(
+            "Installed versions",
+            package.versions.keys().map(|k| k.to_string()).collect::<Vec<String>>().join(" | "),
```
BraveChicken1: Maybe it would be nice to implement a display extension for iterators:
`fn display(&self, separator: &str) -> String`
This function would then map the items to a string, collect them and join them together.

src/cli/commands/search.rs
```diff
@@ -63,11 +70,12 @@ impl SearchArgs {
             return;
         }
 
-        print_grid(&matches.into_iter().collect());
+        display::print_grid(&matches.into_iter().map(|p| p.style()).collect());
     }
 
     /// Searches information of a package based on the provided `OptionalPackageId`.
     /// Fails if the given query is not a valid `OptionalPackageId`.
+    /// TODO: Search command should accept an optional id arg instead (similar to the info command)
```
BraveChicken1: The search command already accepts an OptionalPackageId.

src/cli/commands/search.rs
```diff
@@ -63,11 +70,12 @@ impl SearchArgs {
...
     fn search_package(&self) {
         // Get the optional id
         let message = "The given search query isn't a valid `OptionalPackageId`. For regex use `--regex`.";
```
BraveChicken1: This message should probably not include `OptionalPackageId`. It could be changed to `package` or something similar.

src/cli/commands/switch.rs
```diff
@@ -80,9 +82,14 @@ impl HandleCommand for SwitchArgs {
...
-            self.package_name, self.package_version
+        let styled_message = format!(
+            "Successfully switched {} active version to {}",
+            self.package_name.style(),
+            self.package_version.style()
         )
+        .bold()
+        .green();
+
+        println!("{styled_message}");
```
BraveChicken1:
```suggestion
        let styled_message = format!(
            "Successfully switched {} active version to {}",
            self.package_name.style(),
            self.package_version.style()
);

        println!("{}", styled_message.bold().green());
```

src/cli/commands/update.rs
```diff
@@ -111,19 +106,21 @@ impl HandleCommand for UpdateArgs {
...
                     continue;
                 },
             };
 
             match new_package_id {
                 Some(new_package_id) => {
-                    println!("Successfully updated {} to {new_package_id}", optional_id);
+                    let styled_message =
+                        format!("Successfully updated {} to {}", optional_id.style(), new_package_id.style()).bold().green();
+                    println!("{styled_message}");
```
```suggestion
                    let styled_message = format!("Successfully updated {} to {}", optional_id.style(), new_package_id.style());
                    println!("{}", styled_message.bold().green());
```

src/cli/display/aligned_print.rs
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: GPL-3.0-only
+use std::fmt::Display;
+
+pub const VERTICAL_LINE_PREFIX: &str = " \u{2502} ";
+
+pub struct PairAligner {
+    pairs: Vec<(String, String)>,
+}
```
BraveChicken1, line 7: Please add doc comments to the struct and its methods.

src/cli/display/aligned_print.rs
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: GPL-3.0-only
+use std::fmt::Display;
+
+pub const VERTICAL_LINE_PREFIX: &str = " \u{2502} ";
```
BraveChicken1: This constant could be defined in the PairAligner impl. It would then be available through PairAligner::VERTICAL_LINE_PREFIX.

src/cli/display/aligned_print.rs
```diff
@@ -0,0 +1,38 @@
...
+}
+
+impl PairAligner {
+    pub fn new() -> Self {
+        Self { pairs: Vec::new() }
+    }
+
+    #[expect(unused)]
+    pub fn from(pairs: Vec<(impl Display, impl Display)>) -> Self {
+        let pairs = pairs.into_iter().map(|p| (p.0.to_string(), p.1.to_string())).collect();
```
BraveChicken1:
```suggestion
        let pairs = pairs.into_iter().map(|(key,value)| (key.to_string(), value.to_string())).collect();
```

src/cli/display/aligned_print.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    pub fn add(&mut self, key: impl Display, value: impl Display) {
+        self.pairs.push((key.to_string(), value.to_string()));
+    }
+
+    /// Prints key value pairs in an aligned format.
+    pub fn display(&self, prefix: &str) {
+        let Some(longest_key) = self.pairs.iter().max_by_key(|s| s.0.len()) else {
+            return;
+        };
```
BraveChicken1:
```suggestion
        let Some(longest_key) = self.pairs.iter().max_by_key(|(key, _)| key.len()) else { return };
```

src/cli/display/standard_print.rs
```diff
@@ -0,0 +1,53 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+use std::fmt::Display;
```
BraveChicken1, line 3:
```suggestion
// SPDX-License-Identifier: GPL-3.0-only
use std::fmt::Display;
```

src/cli/display/standard_print.rs
```diff
@@ -0,0 +1,53 @@
...
+{
+    fn display(&self) -> String {
+        match self {
+            Some(option) => option.to_string(),
+            None => "None".dimmed().to_string(),
+        }
+    }
+}
+
+/// Prints a list with help of `print_list`.
```
BraveChicken1:
```suggestion
/// Prints a list in the format defined in `print_list`.
```

src/cli/display/standard_print.rs
```diff
@@ -0,0 +1,53 @@
...
+/// In case of an empty list "None".dimmed() is printed.
```
BraveChicken1:
```suggestion
/// In case of an empty list "None" is printed.
```

src/cli/display/styled.rs
```diff
@@ -0,0 +1,31 @@
+use crate::installer::types::{OptionalPackageId, PackageId, PackageName, Version};
+use colored::{ColoredString, Colorize};
```
BraveChicken1:
```suggestion
// SPDX-License-Identifier: GPL-3.0-only
use crate::installer::types::{OptionalPackageId, PackageId, PackageName, Version};
use colored::{ColoredString, Colorize};
```

src/cli/display/styled.rs
```diff
@@ -0,0 +1,31 @@
...
+use std::fmt::Display;
+
+pub trait Styled: Display {
+    fn style(&self) -> ColoredString;
```
BraveChicken1, line 12: Please add a doc comment to the style method here.

src/installer/installer.rs
```diff
@@ -353,83 +424,99 @@ impl<'a> Installer<'a> {
...
+        let script_data = ScriptData::new(
+            &script_file,
+            &install_directory,
+            &package_id,
+            self.config,
+            &script_args,
+            self.options.verbose,
+        );
+
+        // Install external files necessary for testing
```
```suggestion
        // Download external files necessary for testing
```

src/integrity/issue.rs
```diff
@@ -114,11 +113,7 @@ impl Display for Issue {
                 writeln!(f, "The following dependencies are missing in the register:")?;
 
                 for (package, missing_dependency) in missing {
-                    let item = format!(
-                        "  - {} missing {}",
-                        package.to_string().bold().blue(),
-                        missing_dependency.to_string().bold().blue()
-                    );
+                    let item = format!("  - {} missing {}", package.style(), missing_dependency.to_string().bold().blue()); // TODO: Dependency styling (also blue?)
```
BraveChicken1: There is still a TODO here.

src/platforms/os.rs
```diff
@@ -90,6 +90,8 @@ impl OsVersion {
     fn get_version() -> Option<Self> {
         use std::process::Command;
 
+        use crate::cli::display::styled::Styled;
```
BraveChicken1: Please move this one to the top of the file.

src/repositories/types/license.rs
```diff
@@ -28,7 +29,7 @@ impl Licenses {
 impl Display for Licenses {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            Licenses::Unknown => write!(f, "Unknown"),
+            Licenses::Unknown => write!(f, "{}", "Unknown".dimmed().to_string()),
```
BraveChicken1: I wouldn't implement the colorize here, I think using the Styled trait would be a better option here.

src/installer/error.rs
```diff
@@ -24,33 +24,33 @@ pub enum InstallerError {
     #[error("Prebuild checksum does not match")]
     ChecksumError,
 
-    #[error("Package '{package_name}' with version '{}' is not installed.", version.as_deref().unwrap_or("any"))]
+    #[error("Package {} with version '{}' is not installed.", package_name.style(), version.as_deref().unwrap_or("any"))]
     PackageNotFound {
-        package_name: String,
+        package_name: PackageName,
         version: Option<String>,
     },
```
BraveChicken1:
```suggestion
    #[error("Package {} with version '{}' is not installed.", package_name.style(), version.map(|v| &v.styled()).unwrap_or("any"))]
PackageNotFound {
        package_name: PackageName,
version: Option<Version>,
},
```

BraveChicken1 (changes requested):

src/builder/builder.rs
Max13245: We actually do need it here, otherwise the styling is lost.

src/cli/commands/config.rs
Max13245: I think his is clearer, as it follows the pattern of the stuff above.

src/cli/commands/config.rs
Max13245: Same as above.

src/cli/commands/search.rs
Max13245: I'll remove it, as I'll be working on it on another branch. Also the comment isn't correct like you said. I meant to say that the functionallity for only a package name should be separated from the package version, like in the info command. Then people can also see which versions are available for example.

src/integrity/issue.rs
Max13245: I'll delete it and the colored text for dependencies (to avoid too much different things being blue).

src/cli/commands/info.rs
Max13245: I looked at this, at the moment there is only one real use case for this and I would also need two separate methods (one to do this for styled text and one for normal text):

```
pub trait DisplayJoined {
    fn display(self, separator: &str) -> String;
}

impl<T, U: Iterator<Item = T>> DisplayJoined for U
where
    T: Display,
{
    fn display(self, separator: &str) -> String {
        self.map(|p| p.to_string()).collect::<Vec<String>>().join(separator)
    }
}

pub trait DisplayJoinedStyled {
    fn display_styled(self, separator: &str) -> String;
}

impl<T, U: Iterator<Item = T>> DisplayJoinedStyled for U
where
    T: Styled,
{
    fn display_styled(self, separator: &str) -> String {
        self.map(|p| p.style().to_string()).collect::<Vec<String>>().join(separator)
    }
}
```

Max13245: I don't think it's still worth it, unless in the future there will be a lot more of these, which I doubt. What do you think?

src/installer/error.rs
Max13245: It would have to become something like this:
```
version.clone().map(|v| v.style()).unwrap_or("any".normal()
```

Max13245: instead of .clone() we can also use .as_ref()

Max13245: I think this is the best version:
`version.as_ref().map_or("any".normal(), |v| v.style())`

src/cli/commands/info.rs
BraveChicken1: Another option here would be to implement a method that maps all items in the iterator to styled and a separate method that joins an iterator to a string. That would have more use cases I think.

src/installer/error.rs
BraveChicken1: Yes, that looks good. You probably don't even need the .normal() here.

Max13245: We do, both values need to have the same type.

src/builder/builder.rs
BraveChicken1: For me it works fine with just &str

Max13245: Are you sure? For me the link is colored, but not the package name.

BraveChicken1 (approved):

PR #121, opened 2026-07-16
Add build and test requirements to metadata

BraveChicken1:
This PR adds the `build_requirements` and `test_requirements` fields to the metadata. The current only available requirement is `msvc`, detecting the Microsoft Visual C++ toolchain and adding required paths to the environment.

### DIFF
--- a/src/builder/build_env.rs
+++ b/src/builder/build_env.rs
@@ -1,9 +1,13 @@
 // SPDX-License-Identifier: GPL-3.0-only
 use std::{collections::HashMap, path::PathBuf};
 
+use thiserror::Error;
+
 use crate::{
     cli::display::{logging::warning, styled::Styled},
+    platforms::tool_detection::{self, error::ToolDetectionError},
     register::{installed_package_version::InstalledPackageVersion, package_register::PackageRegister},
+    repositories::types::Requirement,
     utils::env::Environment,
 };
 
@@ -27,18 +31,39 @@ const PATH_SEPARATOR: &str = ":";
 #[cfg(target_os = "windows")]
 const PATH_SEPARATOR: &str = ";";
 
+/// The errors that occur during creating the build environment.
+#[derive(Error, Debug)]
+pub enum BuildEnvError {
+    #[error("Cannot add {} to build env {variable}: cannot convert PathBuf to string", path.display())]
+    PathBufConversion {
+        path: PathBuf,
+        variable: String,
+    },
+
+    #[error("Cannot find tool {0} on the system")]
+    ToolNotFound(String),
+
+    #[error("Error while detecting tool on the system")]
+    ToolDetectionError(#[from] ToolDetectionError),
+}
+
+pub type Result<T> = core::result::Result<T, BuildEnvError>;
+
 /// Holds all the data necessary to build a normalized build environment.
 pub struct BuildEnv<'a> {
     prefix_directory: &'a PathBuf,
     dependencies: &'a Vec<&'a InstalledPackageVersion>,
     build_dependencies: Vec<&'a InstalledPackageVersion>,
+    build_requirements: &'a Vec<Requirement>,
     register: &'a PackageRegister,
 }
 
 #[expect(clippy::from_over_into)]
-impl<'a> Into<Environment> for BuildEnv<'a> {
+impl<'a> TryInto<Environment> for BuildEnv<'a> {
+    type Error = BuildEnvError;
+
     /// Converts the `BuildEnv` struct into a normalized `Environment` struct.
-    fn into(self) -> Environment {
+    fn try_into(self) -> Result<Environment> {
         let mut env = Environment::new();
 
         // TODO: maybe also sandbox TMPDIR variable
@@ -65,6 +90,9 @@ impl<'a> Into<Environment> for BuildEnv<'a> {
             };
         }
 
+        // Add requirement specific vars to the build env
+        env.insert_vars(self.create_requirements_vars()?);
+
         // Add macos specific environment variables
         #[cfg(target_os = "macos")]
         {
@@ -74,7 +102,7 @@ impl<'a> Into<Environment> for BuildEnv<'a> {
             // TODO: add xcode paths
         }
 
-        env
+        Ok(env)
     }
 }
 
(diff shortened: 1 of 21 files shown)


src/builder/build_env.rs
```diff
@@ -244,29 +236,60 @@ impl<'a> BuildEnv<'a> {
...
-            Some(path) => parts.push(path.into()),
-            None => warning!("Cannot add Packit prefix directory to build env ACLOCAL_PATH: cannot convert PathBuf to string"),
-        };
+        Ok(environment)
+    }
+}
 
-        parts.join(PATH_SEPARATOR)
+/// Converts a `PathBuf` to a string.
+/// Returns an `Err` if the path cannot be converted.
```
Max13245:
```suggestion
/// Returns a `BuildEnvError::PathBufConversion` if the path cannot be converted.
```

src/integrity/verifier/package.rs
```diff
@@ -636,14 +636,29 @@ fn check_package_test(package_id: &PackageId, register: &PackageRegister, config
...
+        if !requirement.is_satisfied()? {
+            warning!(
+                "Skipping {} test, because requirement '{requirement}' is not satisfied.\n{}",
+                package_id.style(),
+                requirement.get_not_satisfied_message()
+            );
+            return Ok(false);
+        }
+    }
+
```
Max13245: This needs to happen before every test script run, maybe put it in the `run_test_script` method to avoid repetition.

src/platforms/tool_detection/tools.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub fn get_vs_path(&self) -> &PathBuf {
+        &self.vs_path
+    }
+
+    /// Gets the version of the MSVC toolchain.
+    pub fn get_version(&self) -> &Version {
+        &self.version
+    }
+
+    /// Gets the path of the vcvarsall.bat script.
```
Max13245:
```suggestion
    /// Gets the path of the `vcvarsall.bat` script.
```

src/platforms/tool_detection/tools.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub fn get_vcvarsall_path(&self) -> PathBuf {
+        self.vs_path.join("VC").join("Auxiliary").join("Build").join("vcvarsall.bat")
+    }
+
+    /// Gets the arch needed for the vcvarsall.bat execution.
```
Max13245:
```suggestion
    /// Gets the arch needed for the `vcvarsall.bat` execution.
```

src/platforms/tool_detection/windows.rs
```diff
@@ -0,0 +1,53 @@
...
+    if !version_path.exists() {
+        debug!("MSVC version cannot be read from '{}'", version_path.display());
+        return Ok(None);
+    }
+    let version_str = fs::read_to_string(&version_path).err_with_path("read", version_path)?;
+    let version = Version::from_str(version_str.trim())?;
+
+    let msvc = Msvc::new(vs_path, version);
+
+    // Check if vcvarsall.bat exists
```
Max13245:
```suggestion
    // Check if `vcvarsall.bat` exists
```

src/platforms/tool_detection/windows.rs
```diff
@@ -0,0 +1,53 @@
...
+    let vcvarsall = msvc.get_vcvarsall_path();
+    if !vcvarsall.exists() {
+        debug!("vcvarsall.bat script does not exist at '{}'", vcvarsall.display());
```
Max13245:
```suggestion
        debug!("'vcvarsall.bat' script does not exist at '{}'", vcvarsall.display());
```

src/repositories/types/requirement.rs
```diff
@@ -0,0 +1,76 @@
...
+    where
+        S: serde::Serializer,
+    {
+        serializer.collect_str(&self.to_string())
+    }
+}
+
+impl Requirement {
+    /// Checks if a requirement is satisfied.
+    /// Returns true if the requirement is satisfied, false otherwise
```
Max13245:
```suggestion
    /// Returns true if the requirement is satisfied, false otherwise.
```

src/repositories/types/requirement.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub fn is_satisfied(&self) -> tool_detection::error::Result<bool> {
+        match self {
+            Requirement::Msvc => Ok(tool_detection::detect_msvc()?.is_some()),
+        }
+    }
+
+    /// Gets the message that should be shown when the requirement is not satisfied.
+    pub fn get_not_satisfied_message(&self) -> &str {
+        match self {
+            Requirement::Msvc => "Microsoft Visual C++ cannot be found, please install it first.",
```
Max13245:
```suggestion
            Requirement::Msvc => "Microsoft Visual C++ cannot be found, please manually install it first.",
```

src/utils/env.rs
```diff
@@ -40,4 +40,13 @@ impl Environment {
...
     }
+
+    /// Expands the environment with another environment.
+    pub fn expand(&mut self, other_environment: Environment) {
+        self.insert_vars(other_environment.env_vars);
+
+        for stripped_var in other_environment.stripped_vars {
+            self.strip_var(stripped_var);
+        }
+    }
```
Max13245, line 51: I think this might potentially cause problems because of insertion and strip order. If a `strip_var` was added to `other_environment` before an identical `env_vars` was inserted, then inserting `env_vars` before `strip_var` here would result in a different environment.

Max13245 (changes requested):

src/integrity/verifier/package.rs
BraveChicken1: I could move it, but then this should become an error from the `run_test_script` method. The warning would then be printed from the method calling the `run_test_script`.

src/utils/env.rs
BraveChicken1: This shouldn't be an issue since we call the insert and strip functions. These functions ensure that a var is present in only one of the two lists, not in both. See the `insert_vars` and `strip_var` functions for reference.

src/repositories/types/requirement.rs
BraveChicken1: I changed it to `Microsoft Visual C++ cannot be found, please install it from the Microsoft website first`, that's probably even clearer.

src/integrity/verifier/package.rs
Max13245: Mhmm, the warning is different, I hadn't noticed that. Although it might still be a bit safer this way.

BraveChicken1: Yes, I think returning an error is still neat. The caller can handle it as needed. I will change it!

Max13245 (approved):

PR #122, opened 2026-07-19
Refactor prebuild repositories

BraveChicken1:
This PR refactors the prebuild repositories implementation and adds functionality for future use.
- Prebuilds now have a `prebuild_id` instead of the target architecture, this ensures future compatibility with other target specifications.
- Prebuilds now use a toml metadata file instead of the checksum file, which also contains the size of the prebuild.

### DIFF
--- a/src/installer/install_tree.rs
+++ b/src/installer/install_tree.rs
@@ -239,7 +239,7 @@ impl<'a> InstallTreeBuilder<'a> {
         // Check if a prebuild for the package is available
         let revision = install_meta.version_metadata.get_revision_count();
         let prebuild_id = Target::current().architecture.to_string();
-        match self.repository_manager.get_prebuild_url(&install_meta.repository_id, package_id, revision, &prebuild_id) {
+        match self.repository_manager.get_prebuild_meta(&install_meta.repository_id, package_id, revision, &prebuild_id) {
             Ok(Some(_)) => return Ok(None),
             Ok(None) | Err(RepositoryError::RepositoryNotFoundError { .. }) => {},
             Err(e) => error!(e),
--- a/src/installer/install_tree.rs
+++ b/src/installer/install_tree.rs
@@ -238,7 +238,8 @@ impl<'a> InstallTreeBuilder<'a> {
 
         // Check if a prebuild for the package is available
         let revision = install_meta.version_metadata.get_revision_count();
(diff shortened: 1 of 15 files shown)


src/installer/install_tree.rs
```diff
@@ -238,7 +238,8 @@ impl<'a> InstallTreeBuilder<'a> {
 
         // Check if a prebuild for the package is available
         let revision = install_meta.version_metadata.get_revision_count();
-        match self.repository_manager.get_prebuild_url(&install_meta.repository_id, package_id, revision, &Target::current()) {
+        let prebuild_id = Target::current().architecture.to_string();
```
Max13245, line 241: In the future we would get the prebuild_id from the repo metadata right?

src/installer/installer.rs
```diff
@@ -314,8 +315,8 @@ impl<'a> Installer<'a> {
         let calculated_checksum = Checksum::from_bytes(&bytes);
 
         // Check equality of checksum
-        match checksum {
-            Some(checksum) if checksum == calculated_checksum => debug!("{} prebuild checksum matches", package.style()),
+        match prebuild_meta {
+            Some(prebuild_meta) if prebuild_meta.checksum == calculated_checksum => debug!("{} prebuild checksum matches", package.style()),
             _ => return Err(InstallerError::ChecksumError),
```
Max13245: Maybe adjust this error, now this will also be thrown when there is no prebuild metadata.

Max13245 (changes requested):

src/installer/install_tree.rs
BraveChicken1: Yes, I will implement a more flexible prebuild system for the metadata repository that will provide this prebuild_id.

src/installer/installer.rs
BraveChicken1: Then it should probably be a separate error, describing that the prebuild is not found. I will change this.

Max13245 (commented):
Two small things. Is is maybe better to return the `PrebuildNotFound` from the `get_prebuild_meta`, then the error can be implemented there, because now there are three `PrebuildNotFound` implementations.

Also the new `PrebuildNotFound` doesn't show the `prebuild_id`.

BraveChicken1:
Yes, that's probably better. I will change it.

Max13245 (approved):

PR #123, opened 2026-07-21
Fix option styling

Max13245:
This PR fixes the styling of options. In some cases when an option was `None` it would be dimmed, bold and blue, because `some_option.display().bold().blue()` was used. The new `display_or` fixes this problem, by accepting a styling closure which is applied in case of `Some`.

### DIFF
--- a/src/cli/display/standard_print.rs
+++ b/src/cli/display/standard_print.rs
@@ -1,19 +1,31 @@
 // SPDX-License-Identifier: GPL-3.0-only
-use colored::Colorize;
+use colored::{ColoredString, Colorize};
 use std::fmt::Display;
 
-pub trait DisplayOption {
+pub trait DisplayOption<T: Display> {
     /// Returns the correct string display for an `Option<impl Display>`, dimmed when `None`.
     fn display(&self) -> String;
+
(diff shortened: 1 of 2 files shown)


src/cli/display/standard_print.rs
```diff
@@ -1,19 +1,31 @@
...
 
-pub trait DisplayOption {
+pub trait DisplayOption<T: Display> {
     /// Returns the correct string display for an `Option<impl Display>`, dimmed when `None`.
     fn display(&self) -> String;
+
+    /// Returns the correct string display for an `Option<impl Display>`, dimmed when `None` and styled with the given closure if `Some`.
+    fn display_or<F>(&self, some_style: F) -> String
+    where
+        F: FnOnce(&T) -> ColoredString;
```
BraveChicken1, line 12: I think this could even be display_or_styled with T: Styled, but this is also fine.

BraveChicken1 (approved):

src/cli/display/standard_print.rs
Max13245: True, although this is a bit more flexible.

Issue #125, opened 2026-07-24
Packit claims build script cannot be executed when it does execute it


mygizli04:
It appears Packit prints "Cannot execute build script" if the build script exits with a non-zero status code. This message should probably not be displayed unless the build script cannot be executed outright.

<img width="1172" height="426" alt="Image" src="https://github.com/user-attachments/assets/34d4448a-0347-46c4-a3d6-55f34f5e826e" />

BraveChicken1:
Thank you for pointing out this confusing output. I fixed this issue in #128 and took the opportunity to change some more error messages and improve the last lines of script output.

BraveChicken1:
PR #128 is merged, fix will be in the next release.

PR #128, opened 2026-07-24
Fix error messages to represent errors correctly

BraveChicken1:
This PR fixes error messages that describe the error incorrectly and refactors the output when a script fails.
The script output now shows the last lines of the script output after the error, and the output lines are colored black to improve readability of the error message.

This fixes the issue described in #125.

### DIFF
--- a/src/installer/error.rs
+++ b/src/installer/error.rs
@@ -62,19 +62,19 @@ pub enum InstallerError {
     },
 
     // Wrapped custom errors
-    #[error("Cannot find a repository for installation")]
+    #[error("Cannot fetch package metadata from repository")]
     RepositoryError(#[from] RepositoryError),
 
-    #[error("Cannot info package information")]
+    #[error("Cannot get information from register")]
     RegisterError(#[from] RegisterError),
 
-    #[error("Cannot display installer error")]
+    #[error("Error while displaying message on console")]
     DisplayError(#[from] DisplayError),
 
-    #[error("Cannot execute script")]
+    #[error("Error while executing script")]
     ScriptError(#[from] ScriptError),
 
-    #[error("Cannot execute symlink operation")]
+    #[error("Error while executing symlink operation")]
     SymlinkError(#[from] SymlinkError),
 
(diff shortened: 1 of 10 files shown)


src/installer/scripts.rs
```diff
@@ -63,12 +66,36 @@ pub enum ScriptError {
...
+    // Output script execution message
+    let mut output = String::from("Script executed ");
+    match exit_code {
+        Some(exit_code) => output.push_str(&format!("with status code {exit_code}")),
+        None => output.push_str("without a status code"),
+    }
+
+    // Output last script lines if available
+    if !lines.is_empty() {
+        output.push_str("\n\n");
```
Max13245, line 86: If there are no last lines, then there is no newline after the output string anymore right?

src/installer/error.rs
```diff
@@ -62,19 +62,19 @@ pub enum InstallerError {
...
-    #[error("Cannot display installer error")]
+    #[error("Error while displaying message on console")]
     DisplayError(#[from] DisplayError),
 
-    #[error("Cannot execute script")]
+    #[error("Error while executing script")]
     ScriptError(#[from] ScriptError),
 
-    #[error("Cannot execute symlink operation")]
+    #[error("Error while doing symlink operation")]
```
Max13245:
```suggestion
    #[error("Error while executing symlink operation")]
```

Max13245 (changes requested):

src/installer/scripts.rs
BraveChicken1: Yes, but that is expected behaviour, the display of errors should be only one line (without newlines) in most cases.

Max13245: Ah and thiserror will then make sure there is a newline after the error? So we don't have: 
"Error message.   Something else" on the same line?

BraveChicken1: Yes, when printing the message is printed with a newline.

Max13245 (approved):

PR #132, opened 2026-07-26
Verifier checks for the dependencies directory

Max13245:
This PR adds two new checks (and fixes) and a small bug fix. 

The first check check if any dependency symlinks are missing from the dependencies directory of a package.
The other check checks for invalid files. Invalid files meaning files that shouldn't exist in that location. For now this is only implemented on the dependencies directory, but other directories with "known files" could be added in the future. 

When getting the (initial) check length from the verifier the methods `get_initial_check_length` and `get_check_length` returned the wrong length. They counted on a complete list of checks from the `get_initial_checks` and `get_checks` methods. Which in the future will not be correct anymore, the length should be based on the ordered checks which is created from the "check tree".


src/integrity/repairer/general.rs
```diff
@@ -21,3 +21,17 @@ pub fn fix_stray_directories(strays: HashSet<PathBuf>) -> Result<()> {
 
     Ok(())
 }
+
+/// Removes all the invalid files. Note that symlinks aren't (and should be) traversed.
```
BraveChicken1:
```suggestion
/// Removes all the invalid files. Note that symlinks aren't (and shouldn't be) traversed.
```

src/integrity/repairer/package.rs
```diff
@@ -263,6 +263,28 @@ pub fn fix_invalid_dependents(invalid: Vec<(PackageId, PackageId)>, register: &m
...
 
+/// Fixes the missing or incorrect dependency symlinks (incorrect meaning they point to the wrong file).
+pub fn fix_missing_dir_dependencies(missing: Vec<(PackageId, PackageId)>, config: &Config) -> Result<()> {
+    for (package, dependency) in missing {
+        let dependencies_dir = config.prefix_directory.join("dependencies").join(package.to_string());
+        let symlink_destination = config.prefix_directory.join("packages").join(&dependency.name).join(dependency.version.to_string());
+        let dependency_symlink = dependencies_dir.join(&dependency.name);
+
+        // Remove any existing symlink
+        _ = remove_symlink(&dependency_symlink);
```
BraveChicken1:
```suggestion
        _ = symlink::remove_symlink(&dependency_symlink);
```

src/integrity/repairer/package.rs
```diff
@@ -263,6 +263,28 @@ pub fn fix_invalid_dependents(invalid: Vec<(PackageId, PackageId)>, register: &m
...
+
+        // Make sure that the dependencies directory exists
+        if !fs::exists(&dependencies_dir).err_with_path("check existence of", &dependencies_dir)? {
+            fs::create_dir_all(&dependencies_dir).err_with_path("create dirs", &dependencies_dir)?;
+        }
+
+        // Create the correct symlink
+        create_symlink(&symlink_destination, &dependency_symlink)?;
```
BraveChicken1:
```suggestion
        symlink::create_symlink(&symlink_destination, &dependency_symlink)?;
```

BraveChicken1 (changes requested):
Please also don't forget to update the CHANGELOG and verifier docs.

BraveChicken1 (approved):

PR #133, opened 2026-07-27
Extend Licenses to allow nesting and specifying exceptions

BraveChicken1:
This PR extends the license fields to allow specifying more complex licensing. It now allows nesting of values and specifying exceptions. The display of `Licenses` is changed to be in SPDX format and the metacheck command is updated to include checks for the new format.

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -183,6 +183,20 @@ The following operators are available:
 | `\|`        | Can be used to chain multiple bounds, works as an or operator, for example `3\|5-7`. |
 
 
+### Licenses
+
+The licenses can be specified in a structured format, that is easy to parse. A license field can have the following values:
+- `"<license-name>"`
+- `{ name = "<license-name>", with = ["<exception>"] }`
+- `{ any = [ <license-value> ] }`
+- `{ all = [ <license-value> ] }`
+
+The `any` and `all` fields allow nesting another license value. For example, the following complex license is valid: <br>
+`{ all = [ "License-1", { any = [ { name = "License-2", with = ["Exception-1", "Exception-2"] }, "License-3" ] } ] }`
+
+Note that the `<license-name>` and `<exception>` should be a SPDX Identifier if it is available. See https://spdx.org/licenses/.
+
+
 ### Scripts
 
 The scripts define the specific behaviour to install, uninstall or test a specific package. They can be defined globally for a package, per version or per target. On unix systems the script are written in `sh` and have the `.sh` extension. On Windows the scripts are written in `batch` and have the `.bat` extension.
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -194,7 +194,7 @@ The licenses can be specified in a structured format, that is easy to parse. A l
 The `any` and `all` fields allow nesting another license value. For example, the following complex license is valid: <br>
 `{ all = [ "License-1", { any = [ { name = "License-2", with = ["Exception-1", "Exception-2"] }, "License-3" ] } ] }`
 
-Note that the `<license-name>` and `<exception>` should be a SPDX Identifier if it is available. See https://spdx.org/licenses/.
+Note that the `<license-name>` and `<exception>` should be an SPDX Identifier if it is available. See https://spdx.org/licenses/.
 
 
 ### Scripts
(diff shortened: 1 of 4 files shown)


docs/metadata.md
```diff
@@ -183,6 +183,20 @@ The following operators are available:
...
+The licenses can be specified in a structured format, that is easy to parse. A license field can have the following values:
+- `"<license-name>"`
+- `{ name = "<license-name>", with = ["<exception>"] }`
+- `{ any = [ <license-value> ] }`
+- `{ all = [ <license-value> ] }`
+
+The `any` and `all` fields allow nesting another license value. For example, the following complex license is valid: <br>
+`{ all = [ "License-1", { any = [ { name = "License-2", with = ["Exception-1", "Exception-2"] }, "License-3" ] } ] }`
+
+Note that the `<license-name>` and `<exception>` should be a SPDX Identifier if it is available. See https://spdx.org/licenses/.
```
Max13245:
```suggestion
Note that the `<license-name>` and `<exception>` should be an SPDX Identifier if it is available. See https://spdx.org/licenses/.
```

src/integrity/metadata/check.rs
```diff
@@ -476,23 +476,52 @@ impl MetaCheck {
...
+                if licenses.len() == 1 {
+                    let description = format!("List of licenses from {} contains only a single license", package_id.style());
+                    self.issues.push(MetaIssue::default(description));
+                }
+
+                // Check each license in the list
+                for license in licenses {
+                    self.check_license(license, package_id);
+                }
+                return;
```
Max13245, line 501:
```suggestion

                return;
```

src/integrity/metadata/check.rs
```diff
@@ -476,23 +476,52 @@ impl MetaCheck {
...
+                    self.issues.push(MetaIssue::default(description));
+                }
+
+                // Check if list contains only a single license
+                if licenses.len() == 1 {
+                    let description = format!("List of licenses from {} contains only a single license", package_id.style());
+                    self.issues.push(MetaIssue::default(description));
+                }
+
+                // Check each license in the list
```
Max13245:
```suggestion
                // Recursively check each license in the list
```

src/repositories/types/license.rs
```diff
@@ -9,28 +9,72 @@ pub enum Licenses {
...
+    fn display_impl(&self, f: &mut std::fmt::Formatter<'_>, include_parentheses: bool) -> std::fmt::Result {
+        if include_parentheses && matches!(self, Licenses::All { .. } | Licenses::Any { .. }) {
+            write!(f, "(")?;
+        }
+
+        match self {
+            Licenses::Unknown => write!(f, "Unknown")?,
+            Licenses::Single(license) => write!(f, "{license}")?,
+            Licenses::SingleWithExceptions { name, exceptions } => {
+                write!(f, "{name} WITH ",)?;
```
Max13245:
```suggestion
                write!(f, "{name} WITH ")?;
```

src/repositories/types/license.rs
```diff
@@ -9,28 +9,72 @@ pub enum Licenses {
...
+                let exceptions_str = exceptions.join(", ");
+                match exceptions.len() {
+                    1 => write!(f, "{}", exceptions_str)?,
+                    _ => write!(f, "({})", exceptions_str)?,
```
Max13245:
```suggestion
                    1 => write!(f, "{exceptions_str}")?,
                    _ => write!(f, "({exceptions_str})")?,
```

src/repositories/types/license.rs
```diff
@@ -9,28 +9,72 @@ pub enum Licenses {
...
+                }
+            },
+            Licenses::Any { any } => {
+                for (i, license) in any.iter().enumerate() {
+                    if i > 0 {
+                        write!(f, " OR ")?;
+                    }
+                    license.display_impl(f, true)?;
+                }
```
Max13245, line 56: Is there any way you could get the output of this `license.display_impl(f, true)?;` as a string?
Then you could reduce this to:
`any.iter().map(license.to_string()).join(" OR ")`

Same for the all implementation.

Max13245 (changes requested):

src/repositories/types/license.rs
BraveChicken1: We could create a display to string, but I think this is better as we don't have to allocate multiple strings.

Max13245 (approved):

PR #134, opened 2026-07-27
Add exclude_paths field to PrebuildMeta

BraveChicken1:
This PR adds the `exclude_paths` field to the `prebuilds.toml` file. This field allows the exclude certain paths from packaging of a prebuild.

This is for example useful for `ca-certificates`, the generated `cert.pem` file can now be excluded in the prebuild.


Max13245 (commented):
I think this is a good change, however, by including it in the `prebuilds.toml` you would have to specify targets if you want to exclude certain files. Where normally this might not need to be done. 

This is a bit late, but I'm beginning to doubt the `prebuilds.toml` file, isn't its content better placed in the `package.toml` file? This would also solve the passing of the `PackageMeta`.  Then this would also not be an issue and `package.toml` is pretty empty.

Max13245 (approved):

BraveChicken1:
I think we'll see how this new `prebuilds.toml` file will work out. If it needs changes, we can always implement them and change the format. I will merge this for now.

PR #135, opened 2026-07-27
Small refactors for prebuilds

BraveChicken1:
This PR refactors several small things that use prebuilds:
- The `read_prebuilds_list` method now does not require PackageMeta anymore, it returns an Option and the caller needs to call `PrebuildsList::default`.
- The `prebuilds_url` and `prebuilds_provider` fields in the `repository.toml` file now are used as a suggestion, when adding a repository the user is prompted for using this prebuilds repository.
- The package alteration check is now enabled for packages that were installed from a prebuild.

### DIFF
--- a/src/cli/commands/config.rs
+++ b/src/cli/commands/config.rs
@@ -12,6 +12,7 @@ use crate::{
             aligned_print::PairAligner,
             ask_user,
             logging::{error, warning},
+            standard_print::DisplayOption,
             styled::Styled,
         },
     },
@@ -240,7 +241,7 @@ impl ConfigArgs {
             Some(provider) => provider,
             None => DEFAULT_METADATA_PROVIDER_ID,
         };
-        let repository = Repository::new(url, provider);
+        let mut repository = Repository::new(url, provider);
 
         // Try to connect to the repository
         let provider = provider::create_metadata_provider(&repository).unwrap_or_exit_msg("Unable to connect to repository", 1);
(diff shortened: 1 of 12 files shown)


src/cli/commands/config.rs
```diff
@@ -260,6 +261,21 @@ impl ConfigArgs {
...
+        // Check if repository suggests a prebuild url
+        if let Some(prebuilds_url) = &repo_meta.prebuilds_url {
+            println!("This repository suggests using the following prebuilds repository:");
+            let mut pair_aligner = PairAligner::new();
+            pair_aligner.add("Url", prebuilds_url);
+            pair_aligner.add("Provider", repo_meta.prebuilds_provider.display());
+            pair_aligner.display(PairAligner::VERTICAL_LINE_PREFIX);
+
+            if ask_user("Do you want to add this prebuild repository too?", QuestionResponse::Yes).unwrap_or_exit(1).is_yes() {
+                println!("Adding prebuild repository to config.");
```
```suggestion
            if ask_user("Do you want to add this prebuild repository?", QuestionResponse::Yes).unwrap_or_exit(1).is_yes() {
                println!("Adding prebuild repository to config");
```

Max13245 (commented):
Looks really good, just one issue I can think of. What if the suggested prebuild url changes/if the prebuild url in general changes? The user would still have their old prebuild url in the `Config.toml`.

BraveChicken1:
If the prebuild url changes the user has to change it themselves. We could implement an integrity check that checks if the metadata and prebuild repositories are still working. If the prebuild repository is not working and the suggestion is different, we could fix it by changing it to the new suggestion.

Max13245 (approved):

PR #138, opened 2026-07-28
Remove readonly check from the is_writable method

Max13245:
This PR removed the `readonly` check from the `is_writable` method, because it checks writability based on file attributes instead of ACL's on windows. And on Unix systems it doesn't check if the current user is in a group which has write access.
The `install.bat` now also makes sure not to have any Packit files with the readonly attribute.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Fix MSVC detection not finding VS Build Tools.
 - Fix "check length methods" from the verifier returning wrong length in case of an incomplete check list.
 - Fix portable repositories generating an invalid prebuild repository structure.
-- Fix the writable check using the `readonly` method to check write permissions.
+- Fix writable checks by ignoring the readonly attribute on Windows.
 
 
 ## [v0.0.3](https://github.com/pack-it/packit/compare/0.0.2...0.0.3) - 2026-07-11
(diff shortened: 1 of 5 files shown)


src/platforms/permissions.rs
```diff
@@ -26,15 +26,8 @@ pub fn is_writable(path: &PathBuf) -> Result<bool> {
...
-    let metadata = fs::metadata(path).err_with_path("read metadata of", path)?;
-    let permissions = metadata.permissions();
-
-    // Check if path is read only
-    if permissions.readonly() {
-        return Ok(false);
-    }
-
     // Use platform specific writable checks
+    let metadata = fs::metadata(path).err_with_path("read metadata of", path)?;
```
BraveChicken1: I would move the metadata reading to the platform::is_writable function. Only Unix needs it, so we can fully skip reading the metadata on Windows.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Fix MSVC detection not finding VS Build Tools.
 - Fix "check length methods" from the verifier returning wrong length in case of an incomplete check list.
 - Fix portable repositories generating an invalid prebuild repository structure.
+- Fix the writable check using the `readonly` method to check write permissions.
```
```suggestion
- Fix writable checks by ignoring readonly attribute on Windows.
```

BraveChicken1 (changes requested):

BraveChicken1:
Please also remove the readonly check in the Permissions verifier check (verifier/initial.rs line 40-46)

BraveChicken1 (approved):

Issue #127, opened 2026-07-24
Build script for packit causes truncation/corruption of PATH on Windows


mygizli04:
In `build.bat`, on line 250 the following line is present to allow a user to automatically add packit to their path:

```batch
setx PATH "%PATH%;%PREFIX_DIR%\bin"
```

However, `setx` has a limitation that only allows it to set a value up to 1024 characters. If that limit is exceeded, instead of failing, it will display `WARNING: The data being saved is truncated to 1024 characters.` and truncate the user's PATH to 1024 characters, corrupting it.

The workarounds seem to be using a powershell script to do it (either seperate just to set the PATH or by porting the entire script), another executable (either 3rd party or a subcommand in packit) to set the path without the limitation, or remove the option to automatically set the PATH variable.

Additionally, this approach assumes the user wants to preserve their entire PATH, but the PATH for the current session might contain temporary directories (such as manually specified compiler locations) that the user might not want to preserve, for example the bash equivalent would be:

```bash
PATH=$PATH:/tmp/tmp2/bin/ # temporary PATH assignment
# [packit install command here, assume Y for automatically adding packit to path]
# /tmp/tmp2/bin/ is now permanently in PATH, which is likely unintended behaviour
```

Max13245:
Thanks for pointing this out! It's now fixed in #139 with the use of powershell.

PR #139, opened 2026-07-29
Fix truncated PATH in Windows install

Max13245:
This PR fixes the issues states in #127, namely:
- the setx 1024 character limit
- using the entire current PATH to update the PATH

The `install.bat` now uses a powershell command to add Packit to the user PATH.

### DIFF
--- a/install.bat
+++ b/install.bat
@@ -252,7 +252,7 @@ if %ERRORLEVEL%==0 (
 REM Ask the user if they want to automatically add Packit to their PATH
 call :ask "Y" "Do you wish to automatically add Packit to your user PATH"
 if ERRORLEVEL 1 (
-    powershell -NoProfile -Command "[Environment]::SetEnvironmentVariable('Path', '!ORIGINAL_PATH!' + '%PREFIX_DIR%\bin;', 'User')"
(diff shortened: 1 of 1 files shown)


install.bat
```diff
@@ -247,14 +252,13 @@ if %ERRORLEVEL%==0 (
 REM Ask the user if they want to automatically add Packit to their PATH
 call :ask "Y" "Do you wish to automatically add Packit to your user PATH"
 if ERRORLEVEL 1 (
-    setx PATH "%PATH%;%PREFIX_DIR%\bin"
+    powershell -NoProfile -Command "[Environment]::SetEnvironmentVariable('Path', '!ORIGINAL_PATH!' + '%PREFIX_DIR%\bin;', 'User')"
```
BraveChicken1: Does the Windows PATH always have a trailing `;`?

Max13245: Good question, mine did, but I'll check.

Max13245: Not always, I'll fix it.

BraveChicken1 (approved):

BraveChicken1 (approved):

PR #140, opened 2026-07-31
Add skip test flags

Max13245:
This PR adds two new flags to the `install` command, `--skip-test` to skip a Packit test and `--skip-build-test` to skip build script tests.

### DIFF
--- a/src/builder/builder.rs
+++ b/src/builder/builder.rs
@@ -35,16 +35,24 @@ pub struct Builder<'a> {
     register: &'a mut PackageRegister,
     repository_manager: &'a RepositoryManager<'a>,
     verbose: bool,
+    skip_build_test: bool,
 }
 
 impl<'a> Builder<'a> {
     /// Creates new builder
-    pub fn new(config: &'a Config, register: &'a mut PackageRegister, repository_manager: &'a RepositoryManager, verbose: bool) -> Self {
+    pub fn new(
+        config: &'a Config,
+        register: &'a mut PackageRegister,
+        repository_manager: &'a RepositoryManager,
+        verbose: bool,
+        skip_build_test: bool,
+    ) -> Self {
         Self {
             config,
             register,
             repository_manager,
             verbose,
+            skip_build_test,
         }
     }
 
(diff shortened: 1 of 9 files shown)


src/builder/builder.rs
```diff
@@ -35,16 +35,24 @@ pub struct Builder<'a> {
...
 impl<'a> Builder<'a> {
     /// Creates new builder
-    pub fn new(config: &'a Config, register: &'a mut PackageRegister, repository_manager: &'a RepositoryManager, verbose: bool) -> Self {
+    pub fn new(
+        config: &'a Config,
+        register: &'a mut PackageRegister,
+        repository_manager: &'a RepositoryManager,
+        verbose: bool,
+        skip_build_test: bool,
+    ) -> Self {
```
BraveChicken1, line 49: Maybe the builder also needs `BuilderOptions`, like the `InstallerOptions`.

src/cli/commands/install.rs
```diff
@@ -93,12 +106,18 @@ impl HandleCommand for InstallArgs {
             InstallType::Prebuild
         };
 
+        if self.skip_test {
+            warning!("Packit tests are skipped, because the `--skip-test` flag is enabled");
+        }
```
BraveChicken1, line 111: Shouldn't we also show a warning for build tests?

src/installer/installer.rs
```diff
@@ -232,7 +238,10 @@ impl<'a> Installer<'a> {
 
         self.determine_active(install_meta, &package_id, target)?;
 
-        self.execute_test(&package_id, install_meta, &install_directory, &script_args, target)?;
+        // Skip the test if the skip test option is true
```
```suggestion
        // Only run the test if the skip test option is false
```

src/installer/options.rs
```diff
@@ -48,8 +52,21 @@ impl InstallerOptions {
         self
     }
 
+    /// Sets the verbose field.
```
BraveChicken1:
```suggestion
    /// Sets the `verbose` field.
```

src/installer/scripts.rs
```diff
@@ -118,6 +119,7 @@ impl<'a> ScriptData<'a> {
         config: &'a Config,
         args: &'a HashMap<&str, &str>,
         verbose: bool,
+        skip_build_test: bool,
```
BraveChicken1: I wouldn't put the `skip_build_test` in the ScriptData, since only the build script needs this.

BraveChicken1 (changes requested):

src/builder/builder.rs
Max13245: mhmm, maybe, but it wouldn't really solve the problem I think, because we would just have one parameter less. As the config, register and repository_manager wouldn't go in there.

BraveChicken1: That is true, but it would probably be a bit neater than it is now. However we could also do this later.

Max13245: Let's do it later, then we can also look at this more in general.

BraveChicken1 (approved):
Please don't forget to update the README, metadata docs (script variable) and CHANGELOG.

PR #141, opened 2026-08-01
Create a '--tree' flag on the search command

Max13245:
This PR adds the `--tree` flag for the search command. Which can be used to show the dependency tree of a package.


src/cli/commands/search.rs
BraveChicken1, line 144: I would keep the regex suggestion here.

BraveChicken1 (changes requested):

BraveChicken1 (approved):

PR #142, opened 2026-08-01
Fix confusing package not found messages for invalid metadata

BraveChicken1:
This PR adds a new `PackageNotFoundReason::InvalidMetadata` to fix confusing error messages when metadata cannot be parsed. Previously this showed that the package could not be found in any repository, followed by a list of versions (including the invalid version)

### DIFF
--- a/src/repositories/error.rs
+++ b/src/repositories/error.rs
@@ -80,7 +80,7 @@ pub enum PackageNotFoundReason {
 
     #[error("the metadata cannot be parsed: {error}")]
     InvalidMetadata {
-        error: toml::de::Error,
+        error: Box<toml::de::Error>,
     },
 
     #[error("package is disabled since {since}{}", reason.as_ref().map(|reason| format!(", with reason '{reason}'")).unwrap_or_default())]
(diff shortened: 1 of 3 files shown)


src/repositories/error.rs
```diff
@@ -126,6 +131,14 @@ impl PackageNotFoundReason {
                     };
                 },
 
+                // If the current primary is `InvalidMetadata`, only use the new one if it is not `NotFound` or `InvalidMetadata`
+                PackageNotFoundReason::InvalidMetadata { .. } => {
+                    match next {
+                        PackageNotFoundReason::NotFound | PackageNotFoundReason::InvalidMetadata { .. } => {},
+                        _ => primary = next.clone(),
+                    };
+                },
```
Max13245, line 136:
```suggestion
                PackageNotFoundReason::InvalidMetadata { .. } => match next {
                    PackageNotFoundReason::NotFound | PackageNotFoundReason::InvalidMetadata { .. } => {},
                    _ => primary = next.clone(),
                },
```

src/repositories/error.rs
```diff
@@ -126,6 +131,14 @@ impl PackageNotFoundReason {
                     };
                 },
```
Max13245: This diff is not so nice, but it's the same suggestion as for `PackageNotFoundReason::InvalidMetadata`:
```suggestion
PackageNotFoundReason::Disabled { since, .. } => match next {
                    PackageNotFoundReason::UnsupportedTarget | PackageNotFoundReason::NotSupported { .. } => primary = next.clone(),
                    PackageNotFoundReason::Disabled { since: since_next, .. } if since_next > since => primary = next.clone(),
                    _ => {},
                },
```

Max13245 (changes requested):
Looks good!

Please solve the clippy issues about the `RepositoryError` being to large. I think we can just turn these warnings off.

src/repositories/error.rs
BraveChicken1: Do you like this formatting? I explicitly used this formatting for extra code clarity

Max13245: Yes, I think it's quite nice. It removes the extra parentheses, and apart from the "match next" being behind the match arm, everything is still the same

Max13245 (approved):

PR #143, opened 2026-08-02
Add packit info command

BraveChicken1:
This PR adds the `pit info` command to show information about the current packit installation. This command can be used to easily share information about the installation for debugging issues.

### DIFF
--- a/src/cli/commands/info.rs
+++ b/src/cli/commands/info.rs
@@ -2,6 +2,7 @@
 use crate::{
     cli::display::{standard_print::DisplayJoined, styled::MapStyled},
     installer::types::PackageName,
+    platforms::{DEFAULT_CONFIG_DIR, OsVersion, Target},
     utils::packit_version::{packit_version, packit_version_name},
 };
 use clap::Args;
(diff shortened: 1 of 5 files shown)


src/cli/commands/info.rs
Max13245: This way the `--tree` flag can only be used when the option is `Some`:
```suggestion
    #[arg(long, default_value = "false", requires = "package")]
```

src/cli/commands/info.rs
```diff
@@ -69,23 +80,95 @@ impl HandleCommand for InfoArgs {
...
+            error!(msg: "Displaying a tree requires package version to be specified.");
+            exit(1);
+        }
+
+        let target = Target::current();
+
+        // Read latest version of packit
+        let manager = RepositoryManager::new(config);
+        let latest_version;
+        match manager.read_package(&PackageName::from_str("packit").expect("Expected 'packit' to be a valid package name")) {
```
Max13245: Maybe not necessarily something for this PR, but it might be nice to either have a method in the `PackageName` which can get the Packit name. Or maybe a macro like the Packit version. Or maybe even changing the Packit version macro into a `PackageId`.

Max13245:
```suggestion
        let latest_version;
        let packit_name = PackageName::from_str("packit").expect("Expected 'packit' to be a valid package name");
        let packit_id = PackageId::new(packit_name, current_packit_version());
        match manager.read_package(&packit_name) {
```

Then later, this:
`println!("{}{}", "packit@".bold().blue(), packit_version!().bold().blue());`
could turn into this:
println!("{}", packit_id.style());

Max13245 (changes requested):
Looks really nice :)

src/cli/commands/info.rs
BraveChicken1: Yes, I was thinking of that. I could make a `PackageName::packit()` that gets the package name "packit".

Max13245 (approved):

PR #144, opened 2026-08-03
Add download progress bars

BraveChicken1:
This PR implements progress bars for downloads where the size is known.

Progress bars are implemented for:
- Source downloads for builds
- Prebuild downloads

### DIFF
--- a/src/builder/builder.rs
+++ b/src/builder/builder.rs
@@ -18,13 +18,13 @@ use crate::{
     installer::{
         install_tree::InstallMeta,
         scripts::{self, ScriptData, ScriptError},
-        types::PackageId,
+        types::{PackageId, PackageName},
         unpack::{ArchiveExtension, unpack},
     },
     register::package_register::PackageRegister,
     repositories::{
         manager::RepositoryManager,
-        types::{Checksum, FileSize, Patch},
+        types::{Checksum, Patch, Source},
     },
     utils::{
         ioerror::IOResultExt,
@@ -118,8 +118,7 @@ impl<'a> Builder<'a> {
         debug!("Source size: {}", source.size);
 
         // Download the build files
-        let description = package_name.style().to_string();
-        let bytes = self.download_file(&source.url, &source.mirrors, &source.checksum, &description, Some(&source.size))?;
+        let bytes = self.download_source(&source, &package_name)?;
 
         // Create temp directory to build in
         let build_directory = TempDir::new().err_operation("create temp dir")?;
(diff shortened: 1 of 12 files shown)


src/builder/builder.rs
```diff
@@ -197,12 +197,28 @@ impl<'a> Builder<'a> {
...
+                        alternative.cyan()
+                    );
+                    spinner.adjust_message(message);
+                }
+            };
+
+            let bytes = self.download_file(&patch.url, &patch.mirrors, &patch.checksum, callback, None)?;
+            spinner.finish();
+            return Ok(bytes);
         }
```
Max13245, line 222: Maybe do an early return here, it might make the indents in the callback a bit less extreme.

src/builder/builder.rs
```diff
@@ -230,12 +246,33 @@ impl<'a> Builder<'a> {
...
 
-    /// Downloads a file from the url, or one of the mirrors. Checks against a checksum and shows a spinner.
-    fn download_file(&self, url: &str, mirrors: &[String], checksum: &Checksum, download_description: &str) -> Result<Bytes> {
-        // Show download spinner
-        let mut spinner = Spinner::new(format!("Downloading {download_description} from '{}'", url.cyan()));
-        spinner.show();
+    /// Downloads the source file of the package. Shows the download progress in a `ProgressBar`.
+    fn download_source(&self, source: &Source, package_name: &PackageName) -> Result<Bytes> {
+        let line_1 = format!("Retrieving {} from '{}'", package_name.style(), source.url.cyan());
+        let message = format!("{line_1}\nDownloading {}", package_name.style());
```
Max13245:
```suggestion
        let retrieve_message = format!("Retrieving {} from '{}'", package_name.style(), source.url.cyan());
        let download_message = format!("{line_1}\nDownloading {}", package_name.style());
```

src/installer/installer.rs
```diff
@@ -310,22 +310,22 @@ impl<'a> Installer<'a> {
...
         })?;
 
-        let (extension, bytes) = self.repository_manager.read_prebuild(repository_id, package, revision, prebuild_id)?;
         let prebuild_meta = self.repository_manager.get_prebuild_meta(repository_id, package, revision, prebuild_id)?;
 
-        // Finish download spinner
-        spinner.finish();
+        // Show download progress bar
+        let message = format!("Downloading {} prebuild from '{}'", package.name.style(), repository_id);
+        let mut progressbar = ProgressBar::new(prebuild_meta.size.0.into(), message);
```
Max13245, line 323: We are now ignore these two requests, they don't have size information, so it might be nice to have a spinner here.

src/repositories/portable_repo.rs
```diff
@@ -325,7 +325,7 @@ impl<'a> PortableRepoCreator<'a> {
             Err(e) => return Err(e.into()),
         };
 
-        let (_, prebuild) = self.repository_manager.read_prebuild(repository_id, package_id, revision, prebuild_id)?;
+        let (_, prebuild) = self.repository_manager.read_prebuild_bytes(repository_id, package_id, revision, prebuild_id)?;
```
Max13245, line 328: Shouldn't there also be a spinner or progress bar here?

src/utils/reading.rs
```diff
@@ -0,0 +1,60 @@
...
+            // Stop reading if end of stream is reached
+            if n == 0 {
+                break;
+            }
+
+            // Add read buffer to final buffer
+            bytes.extend_from_slice(&buffer[..n]);
+
+            // Call progress callback
+            progress(bytes.len());
```
Max13245, line 58: We should watch out for / test for if the progress method doesn't cause too much overhead, because of it's IO interactions.

Max13245 (changes requested):

src/installer/installer.rs
BraveChicken1: That is true, although they should be pretty fast. I think using a spinner before the progress bar would maybe look a bit strange.

src/repositories/portable_repo.rs
BraveChicken1: No, the portable repo command already shows a spinner.

src/utils/reading.rs
BraveChicken1: Yes, that is true. I will clarify this a bit more in the function comment.

src/builder/builder.rs
BraveChicken1: I will do that. I don't really like the callbacks myself, but it seems like the most clean way to implement this code now. We should probably look into centralizing all progress indicators together some time, to avoid cluttering all code with progress bars.

Max13245: Yes, let's do that!

src/installer/installer.rs
Max13245: Mhmm okay

Max13245 (approved):

PR #145, opened 2026-08-04
Add '--latest' flag to search command

Max13245:
This PR adds the `--latest` flag the the search command, so the latest version is used when specified.

### DIFF
--- a/src/cli/commands/search.rs
+++ b/src/cli/commands/search.rs
@@ -67,12 +67,12 @@ impl HandleCommand for SearchArgs {
             return;
         }
 
-        let config = Config::from(&Config::get_default_path()).unwrap_or_exit_msg("Cannot load config", 1);
-        let manager = RepositoryManager::new(&config);
-
         // Get the optional id
         let message = "The given search query isn't a valid package. For regex use `--regex`.";
         let optional_id = OptionalPackageId::from_str(&self.query).unwrap_or_exit_msg(message, 1);
+
+        let config = Config::from(&Config::get_default_path()).unwrap_or_exit_msg("Cannot load config", 1);
+        let manager = RepositoryManager::new(&config);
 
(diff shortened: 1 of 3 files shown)


src/cli/commands/search.rs
```diff
@@ -62,9 +67,41 @@ impl HandleCommand for SearchArgs {
             return;
         }
 
-        match self.tree {
-            true => self.search_tree(),
-            false => self.search(),
+        let config = Config::from(&Config::get_default_path()).unwrap_or_exit_msg("Cannot load config", 1);
+        let manager = RepositoryManager::new(&config);
```
BraveChicken1: I would swap the loading of config and manager with the input checks.

BraveChicken1 (changes requested):
Also don't forget to change the README and CHANGELOG. You can also include the changes to the info command here.

src/cli/commands/search.rs
```diff
@@ -62,9 +67,41 @@ impl HandleCommand for SearchArgs {
...
+        let optional_id = OptionalPackageId::from_str(&self.query).unwrap_or_exit_msg(message, 1);
+
+        let config = Config::from(&Config::get_default_path()).unwrap_or_exit_msg("Cannot load config", 1);
+        let manager = RepositoryManager::new(&config);
+
+        // Check if there is version ambiguity (version and `--latest` specified)
+        if optional_id.version.is_some() && self.latest {
+            error!(msg: "Version is ambiguous, version and `--latest` are both specified");
+            exit(1);
+        }
```
BraveChicken1: I would also swap these two checks, since the check does not need the config and manager yet.

BraveChicken1 (changes requested):

BraveChicken1 (approved):

PR #147, opened 2026-08-07
Extend config command

BraveChicken1:
This PR extends the config command with three new commands:
- `config repositories set-url`
- `config repositories set-prebuilds`
- `config repositories disable-prebuilds`

It also adds the `disable_prebuilds` field to the `config show` command and changes the EditableConfig::set_repository to preserve comments and existing field ordering.

### DIFF
--- a/src/cli/commands/config.rs
+++ b/src/cli/commands/config.rs
@@ -18,7 +18,7 @@ use crate::{
     },
     config::{Config, EditableConfig, Repository},
     register::package_register::PackageRegister,
-    repositories::{manager::RepositoryManager, metadata::DEFAULT_METADATA_PROVIDER_ID, provider},
+    repositories::{manager::RepositoryManager, metadata::DEFAULT_METADATA_PROVIDER_ID, provider, types::RepositoryMeta},
     utils::{packit_version::current_packit_version, unwrap_or_exit::UnwrapOrExit},
 };
 
@@ -69,6 +69,10 @@ pub enum RepositoriesArgs {
 
         /// The optional provider of the new repository, `web` is used as default
         provider: Option<String>,
+
+        /// True if the repository availability checks should be skipped
+        #[arg(long, default_value = "false")]
+        unchecked: bool,
     },
 
     /// Removes a repository from the config
@@ -87,6 +91,10 @@ pub enum RepositoriesArgs {
 
         /// The new provider of the repository, if no value is given, the old value is used
         provider: Option<String>,
+
+        /// True if the repository availability checks should be skipped
+        #[arg(long, default_value = "false")]
+        unchecked: bool,
     },
 
     /// Sets the prebuilds repository url and provider
@@ -127,9 +135,19 @@ impl HandleCommand for ConfigArgs {
             ConfigArgs::SetMultiuser { multiuser } => self.handle_set_multiuser(config, *multiuser),
             ConfigArgs::Repositories(RepositoriesArgs::List) => self.handle_list_repositories(config),
             ConfigArgs::Repositories(RepositoriesArgs::SetRank { new_rank }) => self.handle_set_repositories_rank(config, new_rank),
-            ConfigArgs::Repositories(RepositoriesArgs::Add { id, url, provider }) => self.handle_add_repository(config, id, url, provider),
+            ConfigArgs::Repositories(RepositoriesArgs::Add {
+                id,
+                url,
+                provider,
+                unchecked,
+            }) => self.handle_add_repository(config, id, url, provider, *unchecked),
             ConfigArgs::Repositories(RepositoriesArgs::Remove { id }) => self.handle_remove_repository(config, id),
-            ConfigArgs::Repositories(RepositoriesArgs::SetUrl { id, url, provider }) => self.handle_set_url(config, id, url, provider),
+            ConfigArgs::Repositories(RepositoriesArgs::SetUrl {
+                id,
+                url,
+                provider,
+                unchecked,
+            }) => self.handle_set_url(config, id, url, provider, *unchecked),
             ConfigArgs::Repositories(RepositoriesArgs::SetPrebuilds {
                 id,
                 prebuilds_url,
@@ -277,7 +295,7 @@ impl ConfigArgs {
     }
 
     /// Handles the config repositories add command.
-    fn handle_add_repository(&self, mut config: EditableConfig, id: &str, url: &str, provider: &Option<String>) {
+    fn handle_add_repository(&self, mut config: EditableConfig, id: &str, url: &str, provider: &Option<String>, unchecked: bool) {
         // Check if the config already contains a repository with this id
         if config.get_config().repositories.contains_key(id) {
             error!(msg: "A repository with id '{id}' already exists.");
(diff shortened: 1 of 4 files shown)


src/cli/commands/config.rs
```diff
@@ -76,6 +76,44 @@ pub enum RepositoriesArgs {
...
+        prebuilds_provider: Option<String>,
+    },
+
+    /// Disables or enables the prebuilds of a repository
+    DisablePrebuilds {
+        /// The id of the repository to enable or disable the prebuilds of
+        id: String,
+
+        /// True to disable prebuilds, false to enable
+        #[arg(action = ArgAction::Set)]
```
Max13245, line 118:
```suggestion
        #[arg(long, action = ArgAction::Set)]
```

src/cli/commands/config.rs
```diff
@@ -307,4 +354,82 @@ impl ConfigArgs {
...
+
+    /// Handles the config repositories set-url command.
+    fn handle_set_url(&self, mut config: EditableConfig, id: &str, url: &str, provider: &Option<String>) {
+        // Check if the config even contains this repository
+        let Some(mut repository) = config.get_config().repositories.get(id).cloned() else {
+            error!(msg: "Repository '{id}' does not exist.");
+            exit(1);
+        };
+
+        warning!("Overwriting url: {}, provider: {}", repository.url, repository.provider);
```
Max13245: I don't think this should be a warning, as this is the intended behaviour. I think it should just be a regular print.

Max13245: Maybe you can check if the URL actually exists (like with the meta-data checks)? Although this might slow down a simle set command.

src/cli/commands/config.rs
```diff
@@ -307,4 +354,82 @@ impl ConfigArgs {
...
+        repository.url = url.into();
+        if let Some(provider) = provider {
+            repository.provider = provider.clone();
+        }
+
+        config.set_repository(id, repository);
+
+        config.save_to(&Config::get_default_path()).unwrap_or_exit_msg("Cannot save config file", 1);
+
+        let styled_message = format!("Succesfully set repository url for '{id}' to {url}!").bold().green();
```
```suggestion
        let styled_message = format!("Succesfully set repository url for '{id}' to '{url}'!").bold().green();
```

src/cli/commands/config.rs
```diff
@@ -307,4 +354,82 @@ impl ConfigArgs {
...
+        // Check if a prebuilds repository was already configured
+        if let Some(old_prebuilds_url) = &repository.prebuilds_url {
+            warning!("Repository '{id}' already had a prebuild repository, overwriting url and provider...");
+            warning!(
+                "Old prebuilds url: {old_prebuilds_url}, provider: {}",
+                repository.prebuilds_provider.display()
+            );
+        }
+
+        repository.prebuilds_url = Some(prebuilds_url.into());
```
Max13245, line 432: Same as for the repository URL, maybe check if it exists.

src/cli/commands/config.rs
```diff
@@ -307,4 +354,82 @@ impl ConfigArgs {
...
+        if prebuilds_provider.is_some() {
+            repository.prebuilds_provider = prebuilds_provider.clone();
+        }
+
+        config.set_repository(id, repository);
+
+        config.save_to(&Config::get_default_path()).unwrap_or_exit_msg("Cannot save config file", 1);
+
+        let styled_message = format!("Succesfully set prebuilds repository for '{id}' to {prebuilds_url}!").bold().green();
```
```suggestion
        let styled_message = format!("Succesfully set prebuilds repository for '{id}' to '{prebuilds_url}'!").bold().green();
```

src/cli/commands/config.rs
```diff
@@ -307,4 +354,82 @@ impl ConfigArgs {
...
+        if value && remove_urls {
+            repository.prebuilds_url = None;
+            repository.prebuilds_provider = None;
+        }
+
+        config.set_repository(id, repository);
+
+        config.save_to(&Config::get_default_path()).unwrap_or_exit_msg("Cannot save config file", 1);
+
+        let disabled = if value { "disabled" } else { "enabled" };
```
Max13245: Maybe change this to "mode" or "status", as it's not always disabled.

src/config.rs
```diff
@@ -239,25 +240,62 @@ impl EditableConfig {
...
 
     /// Sets the repository with the given id.
+    /// Tries to preserve comments and field ordering.
     pub fn set_repository(&mut self, id: &str, repository: Repository) {
         // Ensure repositories is a table, not an inline table
         if !self.document.contains_table("repositories") {
             self.document["repositories"] = toml_edit::Table::new().into();
         }
 
+        let repositories = self.document["repositories"].as_table_like_mut().expect("Expected repositories to be a table!");
```
Max13245, line 250: I think this should be a proper error, because users can (and probably will) edit their own `Config.toml`. If they do it incorrectly, they might get this message which is not meant to reach them.

This actually applies to all the uses of `["some_key"]` as well, so maybe this is something for another PR.

src/cli/commands/config.rs
```diff
@@ -307,4 +354,82 @@ impl ConfigArgs {
...
+        };
+
+        repository.disable_prebuilds = value;
+
+        // Remove url if prebuilds disabled and `--remove-urls` flag is enabled
+        if value && remove_urls {
+            repository.prebuilds_url = None;
+            repository.prebuilds_provider = None;
+        }
+
```
Max13245, line 460: Maybe give a warning if the value is false and the url removal is skipped.

src/config.rs
```diff
@@ -239,25 +240,62 @@ impl EditableConfig {
...
+        // If the value is not present, remove the field.
         if let Some(prebuilds_url) = &repository.prebuilds_url {
-            new_value["prebuilds_url"] = prebuilds_url.into();
+            if repository_table.get("prebuilds_url").map(|x| x.as_str()).flatten() != Some(&prebuilds_url) {
+                repository_table.insert("prebuilds_url", prebuilds_url.into());
+            }
+        } else {
+            repository_table.remove("prebuilds_url");
         }
         if let Some(prebuilds_provider) = &repository.prebuilds_provider {
```
Max13245, line 282:
```suggestion
            repository_table.remove("prebuilds_url");
}

if let Some(prebuilds_provider) = &repository.prebuilds_provider {
```

src/config.rs
```diff
@@ -239,25 +240,62 @@ impl EditableConfig {
     }
 
     /// Sets the repository with the given id.
+    /// Tries to preserve comments and field ordering.
     pub fn set_repository(&mut self, id: &str, repository: Repository) {
```
Max13245, line 244: For this function, why is the change prevented if the value already exists. If the value already exists it wouldn't matter if we overwrite it right? Is it maybe to have a bigger chance of preserving the document style?

Max13245 (changes requested):

src/cli/commands/config.rs
BraveChicken1: We don't need long here, since it is not a flag but a positional argument

src/cli/commands/config.rs
BraveChicken1: I can try to request the repository metadata, I think we even do this already in the add repository command

src/cli/commands/config.rs
BraveChicken1: Prebuild urls are harder to check, since prebuild repositories do not have a repository metadata file

src/config.rs
BraveChicken1: Correctness of the fields is ensured by our Config parsing. An EditableConfig needs the Config (which does parser validation using serde) and the Document (which is not validated). If the Config is parsed succesfully, the document is also valid.

src/cli/commands/config.rs
BraveChicken1: Is that neat? Disabling without removing the urls could but used when a user wants to build certain packages without using prebuilds for some time.

src/config.rs
BraveChicken1: Yes, it is indeed to preserve more document styling. Previously we replaced the repository with a new table, removing field ordering and comments that the user provided.

src/cli/commands/config.rs
Max13245: mhmm, true

src/cli/commands/config.rs
Max13245: Nevermind. My train of thought was that if the `--remove-urls` was false then it wouldn't have to be given at all, but false is the default value of course.

src/cli/commands/config.rs
Max13245: Maybe we can check something like a base URL?

Max13245 (dismissed):

PR #148, opened 2026-08-10
Fix wrongly ignored errors in repairer

Max13245:
This PR fixes the wrongly ignored errors in the repairer.


src/integrity/verifier.rs
BraveChicken1: Wouldn't it be better to just merge these functions together? The next_check function now only checks for the initial checks

BraveChicken1 (changes requested):

BraveChicken1 (approved):

PR #152, opened 2026-08-10
Add use_<script> fields to skip script downloads for non-existent scripts

BraveChicken1:
This PR adds the `use_preinstall`, `use_postinstall` and `use_uninstall` fields to the `targets.toml` file. These need to be set to true for scripts to be found by Packit.

Should Packit return an error for scripts that have a `use_<script>` but not the corresponding script?


Max13245:
Yes, I would throw an error, because at the moment this is not the intended behaviour. If we ever decide that the `use_<script>` field can be inferred, then we could change this.

Max13245 (approved):

PR #153, opened 2026-08-11
Fix cargo install in Packit install scripts

Max13245:
This PR fixes the installation of cargo in the Packit install scripts. The install scripts now try to find cargo in it's default install location, so they don't depend on an updated PATH. The user is also not prompted multiple times for the rustup installation, the default is assumed and also removed at the end of the Packit build.

### DIFF
--- a/install.bat
+++ b/install.bat
@@ -107,7 +107,7 @@ if ERRORLEVEL 1 goto cleanup
 
 REM Install Packit to the prefix directory 
 echo Downloading Packit prebuild from `%PREBUILD_URL%`
-curl --proto "=https" -sSfL %PREBUILD_URL% --output packit@%VERSION%-%REVISION%-%TARGET%.tar.gz
+curl --proto "=https" -sSfL "%PREBUILD_URL%" --output packit@%VERSION%-%REVISION%-%TARGET%.tar.gz
 if not ERRORLEVEL 1 (
     tar -xf packit@%VERSION%-%REVISION%-%TARGET%.tar.gz
     if ERRORLEVEL 1 goto cleanup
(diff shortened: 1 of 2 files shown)


install.bat
```diff
@@ -190,21 +192,20 @@ if not ERRORLEVEL 1 (
...
-    if "!RUSTUP_INSTALLED!"==1 (
-        call :ask "Y" "You installed rustup to install Packit. This installation is not registered in Packit. Do you wish to uninstall it"
-        if ERRORLEVEL 1 (
-            echo Uninstalling rustup
-            rustup self uninstall
-            if ERRORLEVEL 1 goto cleanup
-            echo Uninstalling rustup successful
-        )
+    REM Remove rustup after building Packit
+    echo "!RUSTUP_INSTALLED!"
```
BraveChicken1: This echo should probably be removed

install.sh
```diff
@@ -178,10 +178,12 @@ else
         fi
 
         echo "Installing cargo from 'https://sh.rustup.rs'"
-        curl --proto '=https' --tlsv1.2 -sSfL https://sh.rustup.rs | sh
+        RUSTUP_HOME="$HOME/.rustup"
+        CARGO_HOME="$HOME/.cargo"
```
BraveChicken1, line 188: Do we maybe want to check if $HOME is not empty?

BraveChicken1 (changes requested):

install.sh
Max13245: Yes, probably a good idea. I'll do the same for the `%USER_PROFILE%` in `install.bat`.

BraveChicken1 (approved):

PR #154, opened 2026-08-11
Add license file detection to builder

BraveChicken1:
This PR adds automatic license file detection to the builder. It copies the license files to `<package-prefix>/share/licenses/<package-name>`.

Improvements that can still be made:
- Add a field to the metadata to include extra license files or to exclude certain license files.

### DIFF
--- a/src/builder/builder.rs
+++ b/src/builder/builder.rs
@@ -31,11 +31,18 @@ use crate::{
         manager::RepositoryManager,
         types::{Checksum, Patch, Source},
     },
-    utils::{ioerror::IOResultExt, patches, reading::ReadExt, requests},
+    utils::{io, ioerror::IOResultExt, patches, reading::ReadExt, requests},
 };
 
 /// The list of automatically detected license files
-const LICENSE_FILES: &[&str] = &["license", "licence", "copying", "notice", "copyright"];
+const LICENSE_FILE_NAMES: &[&str] = &["license", "licence", "copying", "notice", "copyright"];
+
+/// The list of automatically detected license file extensions
+#[rustfmt::skip]
+const LICENSE_FILE_EXTENSIONS: &[&str] = &[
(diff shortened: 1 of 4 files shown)


src/builder/builder.rs
```diff
@@ -338,4 +346,56 @@ impl<'a> Builder<'a> {
...
+
+                // Check if file name matches license file names
+                for license_file_name in LICENSE_FILES {
+                    if file_name.starts_with(license_file_name) {
+                        found_files = true;
+
+                        // Create destination directory if it does not exist
+                        if !destination_dir.exists() {
+                            fs::create_dir_all(destination_dir).err_with_path("create dirs", destination_dir)?;
+                        }
```
Max13245, line 397: You could move this outside of the loops, because `destination_dir` doesn't change.

Max13245 (changes requested):
Looks good!

Maybe already add the file extension checks?

src/builder/builder.rs
BraveChicken1: That is true, but I don't want to create an empty directory when it's not needed

Max13245: mhmm, yea, fair enough

src/utils/io.rs
```diff
@@ -153,3 +153,14 @@ pub fn normalize_path(path: &Path) -> PathBuf {
 
     path
 }
+
+/// Gets the last extension of the given file name, or None if the file has no extension
+pub fn get_last_extension(file_name: &str) -> Option<&str> {
+    let extension_index = match file_name.chars().rev().position(|x| x == '.') {
+        Some(index) => index,
+        None => return None,
+    };
```
Max13245:
```suggestion
    let extension_index = file_name.chars().rev().position(|x| x == '.')?;
```

Max13245 (changes requested):

Max13245 (approved):

PR #157, opened 2026-08-15
Add license_exclude and license_include fields to source metadata

BraveChicken1:
This PR adds the `license_exclude` and `license_include` fields to the source metadata. These fields can be used to manually control the automatic license file detection mechanism.

### DIFF
--- a/src/builder/builder.rs
+++ b/src/builder/builder.rs
@@ -357,34 +357,7 @@ impl<'a> Builder<'a> {
     /// Only traverse to depth 2, to prevent detecting third party license files.
     fn copy_license_files(&self, build_directory: &Path, destination_dir: &Path, source: &Source) -> Result<()> {
         // Copy all include paths first
-        for include_path_str in &source.license_include {
-            let include_path = build_directory.join(include_path_str);
-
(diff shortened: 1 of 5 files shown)


src/builder/builder.rs
```diff
@@ -355,7 +355,45 @@ impl<'a> Builder<'a> {
     /// Copies license files from the original source into the destination directory.
     /// Does a breadth-first search from the build directory and stops when it finds license files.
     /// Only traverse to depth 2, to prevent detecting third party license files.
-    fn copy_license_files(&self, build_directory: &Path, destination_dir: &Path) -> Result<()> {
+    fn copy_license_files(&self, build_directory: &Path, destination_dir: &Path, source: &Source) -> Result<()> {
```
Max13245, line 358: Maybe split this up into multiple functions?

Max13245, line 358: Maybe collect the paths into a set and then copy them, to avoid "collisions".

Max13245 (changes requested):

Max13245 (approved):

PR #164, opened 2026-08-16
Check for link.exe in install.bat when building with cargo

Max13245:
This PR adds a check in the install.bat which checks if 'link.exe' is available when building with cargo, to make sure that such a build would work.

### DIFF
--- a/install.bat
+++ b/install.bat
@@ -143,6 +143,27 @@ if not ERRORLEVEL 1 (
             goto cleanup
         )
 
+        REM Check if vswhere exists
+        if not exist "C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" (
+            echo Cannot find vswhere to check for 'link.exe' existence
+            goto cleanup
+        )
+
+        if !TARGET!=="aarch64-pc-windows-msvc" (
+            set "MSVC_TARGET=ARM64"
+        ) else (
+            set "MSVC_TARGET=x86.x64"
+        )
+
+        REM Search MSVC build tools using vswhere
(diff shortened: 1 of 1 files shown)


install.bat
```diff
@@ -143,6 +143,25 @@ if not ERRORLEVEL 1 (
             goto cleanup
         )
 
+        if not exist "C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" (
+            echo Cannot find vswhere to check for 'link.exe' existence
+            goto cleanup
+        )
```
BraveChicken1, line 150:
```suggestion

        REM Check if vswhere exists
        if not exist "C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" (
            echo Cannot find vswhere to check for 'link.exe' existence
            goto cleanup
        )
```

install.bat
```diff
@@ -143,6 +143,25 @@ if not ERRORLEVEL 1 (
...
+
+        if !TARGET!=="aarch64-pc-windows-msvc" (
+            set "MSVC_TARGET=ARM64"
+        ) else (
+            set "MSVC_TARGET=x86.x64"
+        )
+
+        for /f "delims=" %%A in ('"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" -latest -property installationPath -products * -requires Microsoft.VisualStudio.Component.VC.Tools.!MSVC_TARGET!') do set "VS_PATH=%%A"
```
BraveChicken1, line 159:
```suggestion

        REM Search MSVC build tools using vswhere
        for /f "delims=" %%A in ('"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" -latest -property installationPath -products * -requires Microsoft.VisualStudio.Component.VC.Tools.!MSVC_TARGET!') do set "VS_PATH=%%A"
```

BraveChicken1 (changes requested):

BraveChicken1 (approved):

PR #170, opened 2026-08-20
Add integration tests

Max13245:
This PR adds integration tests to Packit. An xtask wrapper is used for the setup and cleanup of these tests.


src/utils/constants.rs
```diff
@@ -5,6 +5,12 @@
 use crate::repositories::metadata::WEB_METADATA_PROVIDER_ID;
 
 pub const REGISTER_FILENAME: &str = "Register.toml";
+
+/// Note that when the `test-support` feature flag is enabled `CONFIG_FILENAME` is compiled differently.
+#[cfg(feature = "test-support")]
```
BraveChicken1: Maybe the `test-support` flag could be renamed to `integration-tests`.

xtasks/xtask-test-runner/src/main.rs
```diff
@@ -0,0 +1,222 @@
+use std::{
+    env,
+    fs::{self, File},
+    path::PathBuf,
+    process::{ExitCode, exit},
+};
+use toml::{Value, map::Map};
+
+#[cfg(any(target_os = "macos", target_os = "linux"))]
+pub const DEFAULT_PREFIX: &str = "/opt/packit/test";
```
BraveChicken1, line 11: Do we want to use this prefix? Does this not result in issues from the verifier?

xtasks/xtask-test-runner/src/main.rs
```diff
@@ -0,0 +1,222 @@
...
+impl Drop for Cleanup {
+    fn drop(&mut self) {
+        self.clean();
+    }
+}
+
+/// Wraps around the `cargo test` command to handle the setup and cleanup.
+/// Returns an `ExitCode`.
+fn main() -> ExitCode {
+    // Execute setup, which will partially determine how the cleanup
```
BraveChicken1:
```suggestion
    // Execute setup, which will partially determine how the cleanup is executed
```

xtasks/xtask-test-runner/src/main.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    ExitCode::from(status.code().unwrap_or(1) as u8)
+}
+
+/// Handles the test setup. Creates the `TestConfig.toml` if necessary and a new test prefix.
+/// Returns `Cleanup` containing setup info.
+fn setup() -> Cleanup {
+    let mut existing_config = true;
+    let config_dir = PathBuf::from(DEFAULT_CONFIG_DIR).join("TestConfig.toml");
+    match fs::exists(&config_dir) {
```
BraveChicken1: I think you can just do:
`if config_dir.exists() {`

xtasks/xtask-test-runner/src/main.rs
```diff
@@ -0,0 +1,222 @@
...
+        exit(1);
+    };
+
+    let Some(prefix) = prefix.as_str() else {
+        eprintln!("Invalid string value for 'prefix_directory' in 'TestConfig.toml'");
+        exit(1);
+    };
+
+    let prefix = PathBuf::from(prefix);
+    let register_directory = prefix.join("Register.toml");
```
BraveChicken1:
```suggestion
    let register_file = prefix.join("Register.toml");
```

xtasks/xtask-test-runner/src/main.rs
```diff
@@ -0,0 +1,222 @@
...
+    println!("Using prefix: {}", prefix.display());
+
+    if fs::exists(&prefix).unwrap() {
```
BraveChicken1:
```suggestion
    if prefix.exists() {
```

xtasks/xtask-test-runner/src/main.rs
```diff
@@ -0,0 +1,222 @@
...
+            println!("Test prefix '{}' specified in 'TestConfig.toml' already exists", prefix.display());
+            exit(1);
+        },
+        Ok(_) => {},
+        Err(e) => {
+            eprintln!("Failed to check existence of '{}'", prefix.display());
+            eprintln!("Error: {e}");
+            exit(1);
+        },
+    }
```
BraveChicken1, line 136: This is the same check as above.

xtasks/xtask-test-runner/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "xtask-test-runner"
+description = "Utility that wraps around cargos test framework which handles setup and cleanup"
```
BraveChicken1:
```suggestion
description = "Utility to run Packit integration tests"
```

Cargo.toml
```diff
@@ -63,3 +63,13 @@ windows = { version = "0.62.2", features = [
     "Win32_System_SystemServices",
     "Win32_System_Memory",
 ] }
+
+[dev-dependencies]
+assert_cmd = "2.2.2"
+
+[features]
+test-support = []
```
BraveChicken1:
```suggestion
integration-tests = []
```

BraveChicken1 (changes requested):

xtasks/xtask-test-runner/src/main.rs
Max13245: Why would this result in issues from the verifier?
Also note that this prefix is always cleaned up after the tests, so temporary.

BraveChicken1: I'm not sure if the verifier checks for unknown directories in the prefix, since that could then trigger an issue. But yes indeed, the prefix is temporary, so should not be a problem.

BraveChicken1 (approved):

PR #171, opened 2026-08-23
Refactor and update unit tests

Max13245:
This PR updates and expands the existing unit test modules. Additional unit tests will be part of a next PR as this PR is already quite large.

### DIFF
--- a/src/installer/types/optional_id.rs
+++ b/src/installer/types/optional_id.rs
@@ -91,8 +91,7 @@ impl OptionalPackageId {
 mod tests {
     use crate::installer::types::{
         PackageId, VersionError,
-        package_id::PackageIdError,
-        package_id_tests::create_package_id,
+        package_id::{PackageIdError, tests::create_package_id},
         package_name::{PackageNameError, tests::create_package_name},
         version::tests::create_version,
     };
--- a/src/installer/types/optional_id.rs
+++ b/src/installer/types/optional_id.rs
@@ -103,37 +103,31 @@ mod tests {
         let package_name = create_package_name("test");
         let version = create_version(&[3, 4, 1]);
         let correct_version = PackageId::new(package_name.clone(), version).into();
-        match OptionalPackageId::from_str("test@3.4.1") {
-            Ok(id) => assert_eq!(id, correct_version),
-            Err(e) => panic!("Expected Ok(OptionalPackageId(name: 'test', version: Some(Version(..)))), got Err({e:?})"),
-        }
+        assert_eq!(OptionalPackageId::from_str("test@3.4.1"), Ok(correct_version));
 
         let correct_version = OptionalPackageId {
             name: package_name,
             version: None,
         };
-        match OptionalPackageId::from_str("test") {
-            Ok(id) => assert_eq!(id, correct_version),
-            Err(e) => panic!("Expected Ok(OptionalPackageId(name: 'test', version: None)), got Err({e:?})"),
-        }
+        assert_eq!(OptionalPackageId::from_str("test"), Ok(correct_version));
     }
 
     #[test]
     fn from_str_empty_optional() {
-        assert!(matches!(
+        assert_eq!(
             OptionalPackageId::from_str(""),
             Err(PackageIdError::PackageNameError(PackageNameError::InvalidPackageName))
-        ))
+        )
     }
 
     #[test]
     fn from_str_invalid_chars() {
         let invalid_chars = "!#$%^&*()~:;{}[]<>,.?/|\\\"\'`+=";
         for char in invalid_chars.chars() {
-            assert!(matches!(
+            assert_eq!(
                 OptionalPackageId::from_str(format!("{char}@3.4.1").as_str()),
                 Err(PackageIdError::PackageNameError(PackageNameError::InvalidPackageName))
-            ))
+            )
         }
     }
 }
--- a/src/installer/types/optional_id.rs
+++ b/src/installer/types/optional_id.rs
@@ -89,14 +89,19 @@ impl OptionalPackageId {
 
 #[cfg(test)]
 mod tests {
-    use crate::installer::types::{PackageId, Version, package_id::PackageIdError, package_name::PackageNameError};
+    use crate::installer::types::{
+        PackageId,
+        package_id::PackageIdError,
+        package_name::{PackageNameError, tests::create_package_name},
+        version::tests::create_version,
+    };
 
     use super::*;
 
     #[test]
-    fn from_str_optional() {
-        let package_name = PackageName::from_str("test").expect("Expected valid package name");
-        let version = Version::from_str("3.4.1").expect("Expected Version");
+    fn valid_from_str_optional() {
+        let package_name = create_package_name("test");
+        let version = create_version(&[3, 4, 1]);
         let correct_version = PackageId::new(package_name.clone(), version).into();
         match OptionalPackageId::from_str("test@3.4.1") {
             Ok(id) => assert_eq!(id, correct_version),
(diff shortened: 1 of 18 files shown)


src/installer/types/optional_id.rs
```diff
@@ -79,56 +79,119 @@ impl OptionalPackageId {
...
 }
 
 #[cfg(test)]
 mod tests {
-    use crate::installer::types::{PackageId, Version, package_id::PackageIdError, package_name::PackageNameError};
+    use crate::installer::types::{
+        PackageId, VersionError,
+        package_id::PackageIdError,
+        package_id_tests::create_package_id,
+        package_name::{PackageNameError, tests::create_package_name},
```
BraveChicken1: Why don't we just use package_id::tests here, instead of package_id_tests

src/installer/types/optional_id.rs
```diff
@@ -79,56 +79,119 @@ impl OptionalPackageId {
...
     }
 
     #[test]
     fn from_str_empty_optional() {
-        assert!(matches!(
+        assert_eq!(
             OptionalPackageId::from_str(""),
             Err(PackageIdError::PackageNameError(PackageNameError::InvalidPackageName))
-        ))
+        )
```
BraveChicken1, line 143: I don't really like using `assert_eq`, as it clutters the code with `PartialEq` implementations on errors.

src/installer/types/version.rs
```diff
@@ -155,62 +156,45 @@ impl FromStr for Version {
...
-}
-
-/// Implements the from trait for `&[u32; N]`.
-impl<const N: usize> From<&[u32; N]> for Version {
-    fn from(value: &[u32; N]) -> Self {
-        Self {
-            numbers: value.iter().map(|v| VersionNumber::from(*v)).collect(),
-        }
-    }
-}
```
BraveChicken1, line 174: These should not be removed, since they are still used in the code.

src/installer/types/version.rs
```diff
@@ -155,62 +156,45 @@ impl FromStr for Version {
...
     #[test]
     fn compare() {
-        let version_a = Version::from(&[3, 4, 0]);
-        let version_b = Version::from(&[3, 4, 0]);
-        let version_c = Version::from(&[3, 4, 1]);
-        let version_d = Version::from(&[3, 3, 5]);
+        let version_a = create_version("3.4.0");
+        let version_b = create_version("3.4.0");
+        let version_c = create_version("3.4.1");
+        let version_d = create_version("3.3.5");
```
BraveChicken1, line 226: I'm not sure if we should change this to using the from_str implementation. Using the from with an array seems to safer.

src/installer/types/version_intervals.rs
```diff
@@ -140,75 +140,128 @@ impl VersionIntervals {
...
+    pub fn create_version_intervals(intervals_str: &str) -> VersionIntervals {
+        VersionIntervals::from_str(intervals_str).expect("Expected valid version intervals string")
+    }
+
     #[test]
     fn from_str_ranges() {
         let version_intervals = VersionIntervals::from_str("<6.6|6.7|6.8-7.10|>8");
+        let intervals = match version_intervals {
+            Ok(intervals) => intervals,
+            Err(e) => panic!("Test failed: {e}"),
```
BraveChicken1: This message does not match the format we used in the other spots

src/installer/types/version_intervals.rs
```diff
@@ -140,75 +140,128 @@ impl VersionIntervals {
...
+        );
+        assert_eq!(version_bounds.get(3), Some(&VersionBounds::Higher(create_version("8"))));
     }
 
     #[test]
     fn from_str_ranges_empty() {
         let version_intervals = VersionIntervals::from_str("");
+        let intervals = match version_intervals {
+            Ok(intervals) => intervals,
+            Err(e) => panic!("Test failed: {e}"),
```
BraveChicken1: Same for this message

src/installer/types/version_intervals.rs
```diff
@@ -140,75 +140,128 @@ impl VersionIntervals {
...
         let intervals = ["3", "<6.6|6.7|6.8-7.10|>8", "<=4|4.5|5-6|>=10.1", "4-10", "32|>34", "<6.5|>6.5"];
         for interval in intervals {
-            match VersionIntervals::from_str(interval) {
-                Ok(_) => {},
-                Err(e) => panic!("Expected Ok(..), got Err({e:?})"),
-            }
+            let parsed_interval = VersionIntervals::from_str(interval);
+            assert!(
+                parsed_interval.is_ok(),
+                "Test failed: {}\nTesting {interval:?}",
```
BraveChicken1: And also for this message

src/installer/types/version_number.rs
```diff
@@ -113,27 +115,33 @@ mod tests {
...
+    }
+
+    #[test]
+    fn from_str_illegal_chars() {
+        for invalid_char in "- .@<>=+|!#$%^&*(),'\"\\?[]{}`~±§".chars() {
+            assert_eq!(
+                VersionNumber::from_str(&invalid_char.to_string()),
+                Err(VersionError::IllegalCharacterError)
+            );
         }
```
BraveChicken1: Shouldn't a VersionNumber only consist of numbers? This test could then maybe also cover normal letters

BraveChicken1 (changes requested):

src/installer/types/optional_id.rs
Max13245: I specifically do it because it makes the test output on failure a lot clearer with:
left: some_value
right: some_value_different

Also the `PartialEq` is only implemented in case the tests are run.

src/installer/types/version.rs
Max13245: Where? As far as I could see they aren't used anywhere except for tests.

src/installer/types/version.rs
Max13245: Actually the current `From` implementation has a bug, it doesn't check for empty input. The `from_str` does and is the only method used elsewhere in the code base.

src/installer/types/optional_id.rs
BraveChicken1: If it actually improves the test output, I understand the choice. Lets keep it like that then!

src/installer/types/version.rs
BraveChicken1: The Windows OS check (platforms/os.rs) still uses this to convert the retrieved Windows version into a Version struct.

src/installer/types/version.rs
BraveChicken1: Then we should probably fix that bug, as the From implementation is still used.

Max13245: Alright, I'll change into a `TryFrom` implementation then, because it needs to return an error.

src/installer/types/version_intervals.rs
Max13245: The reason this one deviates is that there are multiple different values being tested, so I wanted to make it clear which test value failed. I could maybe change it into: `"{}\nTest value: {interval:?}",`?

BraveChicken1: Yes, that's fine.

src/platforms/os.rs
```diff
@@ -9,7 +9,6 @@ use crate::{
 #[cfg(any(target_os = "macos", target_os = "linux"))]
 use std::str::FromStr;
 
-#[cfg(any(target_os = "macos", target_os = "linux"))]
 use crate::cli::display::logging::error;
```
BraveChicken1:
```suggestion
use crate::{
cli::display::{logging::{debug, error}, styled::Styled},
installer::types::Version,
};

#[cfg(any(target_os = "macos", target_os = "linux"))]
use std::str::FromStr;
```

BraveChicken1 (changes requested):

BraveChicken1 (approved):

PR #174, opened 2026-08-27
Add ordered uninstalls

Max13245:
This PR adds ordered uninstalls, so dependents are uninstalled before their dependencies. If a package cannot be uninstalled, because packages still depend on it, a nice list of those dependents is shown.

### DIFF
--- a/src/cli/commands/uninstall.rs
+++ b/src/cli/commands/uninstall.rs
@@ -1,16 +1,19 @@
 // SPDX-License-Identifier: GPL-3.0-only
 use clap::Args;
 use colored::Colorize;
-use std::process::exit;
+use std::{collections::VecDeque, process::exit};
 
 use crate::{
     cli::{
         commands::HandleCommand,
         display::{logging::error, not_found, standard_print, styled::Styled},
-        parameter_checks,
+        parameter_checks::{self, contains_package_id},
     },
     config::Config,
-    installer::{Installer, InstallerOptions, types::OptionalPackageId},
+    installer::{
+        Installer, InstallerOptions,
+        types::{OptionalPackageId, PackageId},
+    },
     register::package_register::PackageRegister,
     repositories::manager::RepositoryManager,
     utils::unwrap_or_exit::UnwrapOrExit,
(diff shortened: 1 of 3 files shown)


src/cli/commands/uninstall.rs
```diff
@@ -1,16 +1,19 @@
...
 use clap::Args;
 use colored::Colorize;
-use std::process::exit;
+use std::{collections::VecDeque, process::exit};
 
 use crate::{
     cli::{
         commands::HandleCommand,
         display::{logging::error, not_found, standard_print, styled::Styled},
-        parameter_checks,
```
BraveChicken1, line 10: I would reference to `contains_package_id` by using `parameter_checks::contains_package_id`.

src/cli/commands/uninstall.rs
```diff
@@ -70,3 +66,63 @@ impl HandleCommand for UninstallArgs {
         register.save_to(&register_dir).unwrap_or_exit(1);
     }
 }
+
+impl UninstallArgs {
+    /// Determines the order in which packages need to be uninstalled.
+    /// Shows an error and exists if a package is a dependency and the dependent is not in the given packages.
```
BraveChicken1:
```suggestion
    /// Shows an error and exits if a package is a dependency and the dependent is not in the given packages.
```

src/cli/commands/uninstall.rs
```diff
@@ -70,3 +66,63 @@ impl HandleCommand for UninstallArgs {
...
+            Some(package_id) if let Some(package_version) = register.get_package_version(&package_id) => {
+                return package_version.dependents.iter().cloned().collect();
+            },
+            Some(package_id) => not_found::register_package_version(&package_id, &register),
+            None if let Some(package) = register.get_package(&optional_id.name) => {
+                return package.get_versions().iter().flat_map(|p| p.dependents.iter().cloned()).collect();
+            },
+            None => not_found::register_package(&optional_id.name, &register),
+        }
+    }
```
BraveChicken1: I would not do the `if let Some` here, I think it overcomplicates the code here. I would put the statement inside the match arm body. You could even do `let Some(package_id) = optional_id.versioned()` instead of the match statement.

src/cli/parameter_checks.rs
```diff
@@ -73,3 +73,18 @@ pub fn expand_optional_ids(register: &PackageRegister, config: &Config, packages
 
     package_ids
 }
+
+/// Checks if a given list of `OptionalPackageId`s contains a `PackageId`.
+/// The list contains the given package id if any of the optional ids match with the package id.
```
BraveChicken1:
```suggestion
/// Returns true if any of the `packages` matches `package_id`, false otherwise.
```

src/cli/commands/uninstall.rs
```diff
@@ -70,3 +66,63 @@ impl HandleCommand for UninstallArgs {
...
+                        msg: "{} cannot be uninstalled, because it is a dependency of the following packages:",
+                        package.style()
+                    );
+                    standard_print::print_list(dependents.into_iter().map(|d| d.style()));
+                    exit(1);
+                }
+            }
+
+            // Note that because we know that the dependent is in the list of given packages it will be before the current package
+            // if we add it to the back of the queue.
```
BraveChicken1:
```suggestion
            // Note that because we know that the dependent is in the list of given packages
            // it will be before the current package if we add it to the back of the queue.
```

BraveChicken1 (changes requested):

BraveChicken1 (approved):

PR #175, opened 2026-08-28
Add --exclude for util package command

Max13245:
This PR adds the `--exclude` flag to the package command. It already exists in the `update` command, so it seems nice to have similar behaviour here. When `--all` is specified `--exclude` specifies packages which should be excluded from the list of packages from `--all`.

### DIFF
--- a/src/cli/commands/util/package.rs
+++ b/src/cli/commands/util/package.rs
@@ -10,9 +10,10 @@ use crate::{
     cli::{
         commands::HandleCommand,
         display::{Spinner, logging::error, not_found, styled::Styled},
+        parameter_checks,
     },
     config::{Config, Repository},
-    installer::types::PackageId,
+    installer::types::{OptionalPackageId, PackageId},
     packager,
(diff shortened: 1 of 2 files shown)


src/cli/commands/util/package.rs
```diff
@@ -34,8 +35,13 @@ pub struct PackageArgs {
     structured: bool,
 
     /// True to package all installed packages
-    #[arg(short, long, default_value = "false")]
+    #[arg(short, long, default_value = "false", conflicts_with = "packages")]
     all: bool,
+
+    /// Exclude packages when using the `--all` flag, specified with <PACKAGE-NAME> ...
+    /// Note that `num_args = 1..` is needed to consume all 'non-flag' values after the exclude
```
BraveChicken1: I would not put this as a doc comment, because clap will then show it in the help menu

BraveChicken1 (changes requested):

BraveChicken1 (approved):

PR #176, opened 2026-08-28
Expand documentation

Max13245:
This PR expands and improves the current documentation.


docs/commands/check.md
```diff
@@ -0,0 +1,27 @@
...
+The `check` command has the following command line syntax:<br>
+`pit check [<PACKAGE-NAME>[@<VERSION>] ...]`
+
+## Basic check
+The `check` command checks the specified packages, using the following syntax:<br>
+`pit check [<PACKAGE-NAME>[@<VERSION>] ...]`
+
+The issues will be listed according to urgency, the first issue being the most urgent. The most urgent issue needs to be fixed first, as it could be causing the other issues.
+
+More information about the workings of this command can be found at [verifier](../verifier.md). If you also want Packit to fix the issues for you check out the [fix](./fix.md) command.
```
BraveChicken1:
```suggestion
More information about the workings of this command can be found at [verifier](../verifier.md). If you also want Packit to fix the issues for you, take a look at the [fix](./fix.md) command.
```

docs/commands/search.md
```diff
@@ -0,0 +1,35 @@
+# Search
+
+The `search` command has the following command line syntax:<br>
+`pit search <QUERY> [--regex] [--verbose | -v] [--tree] [--latest] [--target-only]`
+
+## Basic search
+The `search` command searches for a package, using the following syntax:<br>
+`pit search <QUERY>`
+
+The command shows information based on the package metadata. The `<QUERY>` needs to be a package name or a package name with a version. If the version is given that specific version is searched for.
```
BraveChicken1: This should probably also already describe the `--regex` option, as this line says the query can only be a package

docs/commands/switch.md
```diff
@@ -0,0 +1,22 @@
...
+
+### Examples
+To switch the version of a package called `foo`, use:<br>
+`pit switch foo 1.3.8`
+
+## Flags
+This is a complete list of all flags that can be used with the `pit switch` command.
+
+### `--skip-symlinking`
+Packit packages are symlinked, to learn more about this see [symlinks](../structure.md#symlinks). To skip symlinking `--skip-symlinking` can be used.
```
BraveChicken1:
```suggestion
Active packages are symlinked in most cases, to learn more about this see [symlinks](../structure.md#symlinks). To skip symlinking `--skip-symlinking` can be used.
```

docs/meta-checks.md
```diff
@@ -0,0 +1,10 @@
+# Metadata checks
+
+The metadata checks perform checks on the metadata. It looks for missing, inconsistent or incomplete data. The issues are collected and then shown at the end of the checks. The following table shows the different issue types and what they mean.
+| Issue Type  | Explanation                                                                                          |
```
BraveChicken1, line 5:
```suggestion
The metadata checks perform checks on the metadata. It looks for missing, inconsistent or incomplete data. The issues are collected and then shown at the end of the checks. The following table shows the different issue types and what they mean.

| Issue Type  | Explanation                                                                                          |
```

docs/meta-checks.md
```diff
@@ -0,0 +1,10 @@
...
+| ----------- | ---------------------------------------------------------------------------------------------------- |
+| `Fatal`     | The issue is so severe that the checks cannot be continued and the command has to exit early.        |
+| `Breaking`  | The metadata checks have found an issue which will break certain logic, but continueing is possible. |
```
BraveChicken1:
```suggestion
| `Breaking`  | The metadata checks have found an issue which will break certain logic, but continuing is possible. |
```

docs/meta-checks.md
```diff
@@ -0,0 +1,10 @@
...
+| `Warning`   | The metadata is correct and functions, however it's unconventional (although maybe unavoidable)      |
+
+In certain cases it's possible to make a good guess about what would be a correct metadata value. In such a this value is shown as a suggestion together with the issue. Occasionally some checks are skipped due to a found issue, if this happens it is shown to you together with the issue.
```
BraveChicken1:
```suggestion
In certain cases it's possible to make a good guess about what would be a correct metadata value. In such case this value is shown as a suggestion together with the issue. Occasionally some checks are skipped due to a found issue, if this happens it is shown to you together with the issue.
```

docs/metadata.md
```diff
@@ -10,82 +12,109 @@ This file should be present in every Packit repository, it quickly describes wha
...
 | `apply_patches_in` | Defines the directory that should be used to apply all patches in.                                           |
 | `patches`          | A list of patches to apply to the source. See patches section below.                                         |
 
-#### Patches
+#### Example
+In this example there are multiple sources, because Unix and Windows have different source URLs. Notice that both sources are now named (`unix` and `windows` respectively). If there was only one source this would not be necessary.
+```
+[source.unix]
+url = "https://some-package/unix-version/4.3.tar.gz"
+checksum = "svwergw8e9r6f9w8er6v98we6rv9w8e6rvg9we86rvg0w8ev0werv768669e69vg"
```
BraveChicken1: Please use a correct SHA256 hash here

docs/metadata.md
```diff
@@ -10,82 +12,109 @@ This file should be present in every Packit repository, it quickly describes wha
...
-| `mirrors`  | Defines a list of mirrors which could be used to download the patch if the original url is unavailable. |
-| `apply_in` | Defines the directory that should be used to apply this specific patches in.                            |
-
-Please note that the `mirrors` is not used when the url contains a file in the repository, this file is then expected to exist.
-
-#### Deprecation
+| `mirrors`  | Defines a list of mirrors which could be used to download the patch if the original URL is unavailable. |
+| `apply_in` | Defines the directory that should be used to apply this specific patch in.                              |
+
+Please note that the `mirrors` is not used when the URL contains a file in the repository, this file is then expected to exist.
```
BraveChicken1:
```suggestion
Please note that the `mirrors` is not used when the URL points to a file in the repository, this file is then expected to exist.
```

docs/metadata.md
```diff
@@ -180,16 +196,15 @@ The following operators are available:
...
 | `-`         | Specifies a version range, for example `1-2` (does not include 2).                   |
-| `-=`        | Specifies an including version range, for example `1-2` (does include 2).            |
+| `-=`        | Specifies an including version range, for example `1-=2` (does include 2).           |
 | `<=`        | Specifies a version upper bound including the specified version, for example `<=2`.  |
 | `<`         | Specifies a version upper bound excluding the specified version, for example `<2`.   |
 | `>=`        | Specifies a version lower bound including the specified version, for example `>=1`.  |
 | `>`         | Specifies a version lower bound excluding the specified version, for example `>1`.   |
-| `\|`        | Can be used to chain multiple bounds, works as an or operator, for example `3\|5-7`. |
-
+| `\|`        | Can be used to chain multiple bounds, works as an `OR` operator, for example `3\|5-7`. |
```
BraveChicken1: Please add two spaces to the back of the second row, to match the last item

docs/structure.md
```diff
@@ -0,0 +1,78 @@
...
+| Windows  | `C:\Program Files\packit` |
+
+### Register.toml
+The `Register.toml` file is located inside the prefix and stores information about all installed packages. This file is managed by Packit and should not be changed directly.
+
+### Packages
+All installed packages will go in `<prefix>/packages/<PACKAGE-NAME>/<PACKAGE-VERSION>/`.
+
+### Symlinks
+The following package directories are symlinked in the prefix directory: bin, gnubin, lib, include and share. The gnubin directory exists on macOS to put GNU packages in which conflict with their macOS variants. In such cases a symlink `<prefix>/bin/g<package-name>` → `<prefix>/gnubin/<package-name>` with a 'g' prefix is created, to differentiate between a system and its GNU variant. To use use the GNU tools directly you could add the gnubin directory to your `PATH` as well.
```
BraveChicken1:
```suggestion
The following package directories are symlinked in the prefix directory: bin, gnubin, lib, include and share. The gnubin directory exists on macOS to isolate GNU packages that conflict with their macOS variants. In such case a symlink `<prefix>/bin/g<package-name>` → `<prefix>/gnubin/<package-name>` with a 'g' prefix is created, to differentiate between a system and its GNU variant. To us the GNU tools and overwrite the macOS variants, you could add the gnubin directory to your `PATH` as well.
```

docs/structure.md
```diff
@@ -0,0 +1,78 @@
...
+
+The [active](#active-packages) binaries will be symlinked in `<prefix>/bin/<EXECUTABLE-NAME>`. The `<prefix>/bin` directory needs to be present in the users `PATH` in order for installed binaries to be detected by the system.
+
+### Active packages
+Packit can install multiple versions of a package next to each other. A package always has one active version, which will be symlinked as `<prefix>/active/<PACKAGE-NAME>` → `<prefix>/packages/<PACKAGE-NAME>/<ACTIVE-PACKAGE-VERSION>`
+
+### Dependencies
+The dependencies directory contains symlinks to the dependencies of a package. The structure is as follows `<prefix>/dependencies/<package-name>/<dependency>`, where `<package-name>` is a package with dependency `<dependency>`. `<dependency>` is a symlink to the current active package.
```
BraveChicken1:
```suggestion
The dependencies directory contains symlinks to the dependencies of a package. The structure is as follows `<prefix>/dependencies/<package>/<dependency>`, where `<package>` is a package with dependency `<dependency>`. `<dependency>` is a symlink to the current active package. 
```

docs/structure.md
```diff
@@ -0,0 +1,78 @@
...
+
+The [active](#active-packages) binaries will be symlinked in `<prefix>/bin/<EXECUTABLE-NAME>`. The `<prefix>/bin` directory needs to be present in the users `PATH` in order for installed binaries to be detected by the system.
+
+### Active packages
+Packit can install multiple versions of a package next to each other. A package always has one active version, which will be symlinked as `<prefix>/active/<PACKAGE-NAME>` → `<prefix>/packages/<PACKAGE-NAME>/<ACTIVE-PACKAGE-VERSION>`
+
+### Dependencies
+The dependencies directory contains symlinks to the dependencies of a package. The structure is as follows `<prefix>/dependencies/<package-name>/<dependency>`, where `<package-name>` is a package with dependency `<dependency>`. `<dependency>` is a symlink to the current active package. 
+
+For more detail about the reason why this directory exists see [TODO: a link to some explanation about the patching and how it's used for active ]
```
BraveChicken1: There is still a TODO here

docs/structure.md
```diff
@@ -0,0 +1,78 @@
...
+| Field               | Explanation                                                                                                 |
+| ------------------- | ----------------------------------------------------------------------------------------------------------- |
+| `prefix_directory`  | Defines the directory used for installing packages, see [File structure](#file-structure) for the defaults on each platform. |
+| `repositories_rank` | Defines the order of repositories to search for a package. |
+| `multiuser`         | True to run Packit in multiuser mode, false for single user mode. |
+
+### Repositories
+| Field                 | Explanation                                                                             |
+| --------------------- | --------------------------------------------------------------------------------------- |
+| `url`                 | Defines the url to the repository.                                                      |
```
BraveChicken1:
```suggestion
| `url`                 | Defines the url of the repository.                                                      |
```

src/cli/commands/info.rs
```diff
@@ -133,8 +133,8 @@ impl InfoArgs {
 
         // Create display for new version
         let new_version_disp = match latest_version {
-            Some(latest_version) if latest_version > current_packit_version() => format!("yes, {}", latest_version.style()),
-            Some(_) | None => "no".into(),
+            Some(latest_version) if latest_version > current_packit_version() => format!("{}", latest_version.style()),
+            Some(_) | None => "None".dimmed().to_string(),
```
BraveChicken1, line 137: I would not do this change, I think using "yes, <version>" and "no" is clearer and neater here.

README.md
```diff
@@ -38,170 +38,74 @@ If you only need the `packit` binary itself, you could build it directly using `
...
-If the `--build-all` option is given, the package and all its dependencies are build from source, instead of installing prebuild versions.
-If the `--keep-build` option is given, the build dependencies will not be deleted after building.
-If the `--skip-symlinking` option is enabled, the package is not symlinked into the /bin, /lib, /share, etc. directories.
-If the `--skip-active` option is enabled, the package is not set to active and the current active version is kept. If there is no current active version, this flag is ignored and the package is set to active.
-If the `--skip-test` option is enabled, Packit tests are skipped.
-If the `--skip-build-test` option is enabled, build tests are skipped.
-If the `--verbose` option is given, extra verbose output is shown, like build output.
-If the `--pause-build` option is enabed, the build is paused after build script execution to debug builds.
+#### `pit install <PACKAGE-NAME>[@<VERSION>] ...` [🔗](./docs/commands/install.md)
+Installs the specified packages, if a version is given that version will be installed, if not the latest available version will be installed. Multiple packages can be specified by entering multiple names.
```
```suggestion
Installs the specified packages, if a version is given that version will be installed, if not the latest available version will be installed. Multiple packages can be specified by giving multiple names.
```

README.md
```diff
@@ -38,170 +38,74 @@ If you only need the `packit` binary itself, you could build it directly using `
...
 
-#### `pit uninstall <PACKAGE-NAME>[@<VERSION>] ...`
-Uninstalls the specified packages, if a version is given that version will be uninstalled, if not, you will be asked if you want to delete all versions of `<PACKAGE-NAME>` in case there are multiple versions installed. Multiple packages can be specified by entering multiple names, split by a space.
+#### `pit uninstall <PACKAGE-NAME>[@<VERSION>] ...` [🔗](./docs/commands/uninstall.md)
+Uninstalls the specified packages, if a version is given that version will be uninstalled, if not, you will be asked if you want to delete all versions of `<PACKAGE-NAME>` in case there are multiple versions installed. Multiple packages can be specified by entering multiple names.
```
```suggestion
Uninstalls the specified packages, if a version is given that version will be uninstalled, if not, you will be asked if you want to delete all versions of `<PACKAGE-NAME>` in case there are multiple versions installed. Multiple packages can be specified by giving multiple names.
```

README.md
```diff
@@ -38,170 +38,74 @@ If you only need the `packit` binary itself, you could build it directly using `
...
 
-#### `pit switch <PACKAGE-NAME> <VERSION> [--skip-symlinking]`
-Switches the active version of the specified package to the specified version. If the `--skip-symlinking` option is given, the new active version is not symlinked into the /bin, /lib, /share, etc. directories.
+#### `pit switch-dependency <PACKAGE-NAME>@<VERSION> <DEPENDENCY-NAME> <NEW-DEPENDENCY-VERSION>` [🔗](./docs/commands/switch-dependency.md)
+Switches the dependency (`<DEPENDENCY-NAME>`) version used by a package (`<PACKAGE-NAME>@<VERSION>`) to another version (`NEW-DEPENDENCY-VERSION`).
 
-#### `pit switch-dependency <PACKAGE-NAME>@<VERSION> <DEPENDENCY-NAME> <NEW-DEPENDENCY-VERSION>`
-Switches the dependency version of the specified package to the specified version.
+#### `pit link <PACKAGE-NAME>` [🔗](./docs/commands/link.md)
+The `link` command symlinks the specified package. To learn more about Packit symlinks, check out [symlinks](./docs/structure.md#symlinks).
```
```suggestion
Symlinks the specified package in the prefix directory. To learn more about Packit symlinks, check out [symlinks](./docs/structure.md#symlinks).
```

README.md
```diff
@@ -38,170 +38,74 @@ If you only need the `packit` binary itself, you could build it directly using `
...
-#### `pit config repositories disable-prebuilds <ID> <VALUE> [--remove-urls]`
-Disables or enables the prebuilds url of a repository in the config. If the `--remove-urls` flag is given, the urls are removed if `<VALUE>` is true.
-
-#### `pit init [--prefix <PREFIX>]`
-Initializes the Packit environment by setting up all required files and directories. If the `--prefix` option is given, the given path is used as prefix, instead of the [default prefix](#prefix).
+#### `pit check` [🔗](./docs/commands/check.md)
+Checks the Packit installation for issues.
 
+#### `pit fix` [🔗](./docs/commands/fix.md)
+Fix all issues found by the check command. You will be asked if you want to fix an issue for each issue type.
```
```suggestion
Fixes all issues found by the check command. You will be asked if you want to fix an issue for each issue type.
```

BraveChicken1 (changes requested):

docs/commands/search.md
Max13245: Mhmm, yea, I was hesitating on this one, because it's not part of the "basic" functionality. But I'll reference the `--regex` flag.

src/cli/commands/info.rs
Max13245: I really disagree with that, it's just really messy to use: `true/false`, `yes/no` and `on/off` interchangeably. So, if we have to choose one of the three I would opt for `true/false` as it works for all fields.

README.md
Max13245: Why not 'entering'?

docs/structure.md
Max13245: I don't think it would always overwrite the macOS variants though, because they could be behind them in the `PATH`. So should we state it like this?

BraveChicken1: Then we should probably describe that it needs to be added to the front of the path, otherwise it won't even work.

src/cli/commands/info.rs
BraveChicken1: That's fine for the other changes, but I don't like the `None` here. Maybe then change the name of the field?

README.md
BraveChicken1: Entering sounds more like you enter it in some sort of prompt, but I'm also not fully sure about "giving" though.

Max13245: I think I'll change it to "passing".

src/cli/commands/info.rs
Max13245: Mhmm, I could switch the words around, so:
New available version   None
New available version   3.4.1

docs/build-env.md
```diff
@@ -0,0 +1,46 @@
+# Build Environment
+
+The build environment tries to create an environment which is as 'clean' as possible. This will help with more stable and reproducible builds. This is done by stripping and adding some environmental variables.
```
BraveChicken1:
```suggestion
The build environment tries to create an environment which is as 'clean' as possible. This will help with more stable and reproducible builds. This is done by stripping and adding some environment variables.
```

docs/build-env.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Adjusted PATH
+The build environment creates an adjusted `PATH`. The `PATH` wil contain the bin directories of all (build) dependencies. If on Unix, it will also include the standard Unix system bin paths, which are:
```
BraveChicken1:
```suggestion
The build environment creates an adjusted `PATH`. The `PATH` will contain the bin directories of all (build) dependencies. If on Unix, it will also include the standard Unix system bin paths, which are:
```

docs/build-env.md
```diff
@@ -0,0 +1,46 @@
...
+- `C:\Windows`
+- `C:\Windows\system32`
+- `C:\Windows\System32\Wbem`
+- `C:\Windows\System32\WindowsPowerShell\v1.0`
+
+## Pkgconfig path
+Packit looks for the presence of a `pkgconfig` directory inside of the `share` and `lib` directories of the dependencies of a package. These paths are then added to the `PKG_CONFIG_PATH` to make them available. It also adds `/usr/lib/pkgconfig` which is a macOS specific path.
+
+## Non-symlinked dependencies
+Some packages are not symlinked, so they cannot be found by cmake in the build processs. For that reason, dependencies which are not symlinked are added to the `CMAKE_PREFIX_PATH`. The same applies to the `ACLOCAL_PATH`, with the small difference that build dependencies are also added in this case.
```
BraveChicken1: I think that adding only symlinked packages to the CMAKE_PREFIX_PATH and ACLOCAL_PATH is actually a bug of the build env, I'll change that in a future PR.

docs/build-env.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Requirement environment
+Some requirements need a certain environmental setup. This is different for each requirement. All requirement environments are explained here.
```
BraveChicken1:
```suggestion
Some requirements need a certain environment setup. This is different for each requirement. All requirement environments are explained here.
```

docs/build-env.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Windows MSVC
+Windows MSVC is a requirement for some packages when building on Windows. If MSVC is installed then it is automatically initialized by Packit if necessary. The following environmental variables are then created:
```
BraveChicken1:
```suggestion
Windows MSVC is a requirement for some packages when building on Windows. If MSVC is installed then the installation is automatically detected by Packit if necessary. The following environmental variables are then created:
```

docs/build-env.md
```diff
@@ -0,0 +1,46 @@
...
+| Key                    | Explanation                                                                  |
+| ---------------------- | ---------------------------------------------------------------------------- |
+| `PACKIT_VS_PATH`       | Contains the path to Visual Studio.                                          |
+| `PACKIT_VCVARSALL`     | Contains the path to the `vcvarsall.bat` script.                             |
+| `PACKIT_VCVARSALL_ARCH`| Contains the correct architecture string with which to call `vcvarsall.bat`. |
+| `PACKIT_MSVC_VERSION`  | Contains the MSVC version.                                                   |
+
+The following syntax is used to initialize the MSVC environment in a build script:<br>
```
BraveChicken1:
```suggestion
The following command is required to then initialize the MSVC environment in a build script:<br>
```

docs/build-env.md
```diff
@@ -0,0 +1,46 @@
...
+`call "%PACKIT_VCVARSALL%" %PACKIT_VCVARSALL_ARCH%`
+
+## Extra environmental variables
+In addition to the variables listed above, there are some extra environmental variables. The following variables are added:
```
BraveChicken1:
```suggestion
In addition to the variables listed above, there are some extra environment variables. The following variables are added:
```

docs/installation-process.md
```diff
@@ -0,0 +1,40 @@
...
+| Color   | Explanation                                         |
+| ------- | --------------------------------------------------- |
+| Blue    | Packages which are already installed.               |
+| Cyan    | Packages which are being installed with a prebuild. |
+| Purple  | Packages which are being build from source.         |
+
+This tree is then traversed bottom up, so dependencies are installed first.
+
+## After an installation
+After each package installation the permissions of the package are set. Then the installation of that package is complete and it is saved to the [Register.toml](./structure.md#registertoml). The post-install script is executed after this if it exists. Finally a Packit test script for that package is run to check if the installation was successful. The test script executes some basic functionality of the package.
```
BraveChicken1:
```suggestion
After each package installation the permissions of the package files are set. The installation of that package is then complete and it is added to the [Register.toml](./structure.md#registertoml). The post-install script is executed after this if it exists. Finally a Packit test script for that package is run to check if the installation was successful. The test script executes some basic functionality of the package.
```

docs/installation-process.md
```diff
@@ -0,0 +1,40 @@
...
+
+Once the entire installation is done the build dependencies are removed. It is possible to skip this with use of the [`--keep-build`](./commands/install.md#--keep-build) flag.
+
+## Building from source
+The builder starts by downloading the source files, which it then unpacks into a temporary directory. If [patches](./metadata.md#patches) are specified in the metadata, those will be applied according to their order. After this the [build environment](./build-env.md) is created based on the dependencies and requirements of the package. If [script arguments](./metadata.md#script-environment) are defined in the metadata those will be constructed and added to the environment.
+
+At this point the build script of the package can be run. Afterwards the exit status is checked and returned in case of an error. If[`--pause-build`](./commands/install.md#--pause-build) is specified the builder pauses untill the user chooses to continue.
```
BraveChicken1:
```suggestion
At this point the build script of the package can be run. Afterwards the exit status is checked and returned in case of an error. If[`--pause-build`](./commands/install.md#--pause-build) is specified the builder pauses until the user chooses to continue.
```

docs/installation-process.md
```diff
@@ -0,0 +1,40 @@
...
+
+Now the build is almost done. At this point Packit patches the binaries, not to be confused with the patches defined in the metadata. These patches allow Packit to switch between package dependency versions. A package binary will contain paths to its dependencies. These paths point directly to the dependency package. Packit has a `<prefix>/dependencies/<package-name>` directory which contains symlinks to the different dependency versions. When patching the binaries of a package, Packit searches for paths which have the `<prefix>/packages` prefix. If this is the case, the path should be swapped with the corresponding symlink in the `dependencies` directory.
+
+For example: `<prefix>/packages/foo_dependency/3.7.1` would be changed to `<prefix>/dependencies/foo/foo_dependency`. Where `foo_dependency` point to `<prefix>/packages/foo_dependency/<version>`.
+
+In the end the paths in the binaries point to their dependencies indirectly through the symlinks in the `dependencies` directory.
```
BraveChicken1:
```suggestion
After patching the paths in the binaries point to their dependencies indirectly through the symlinks in the `dependencies` directory.
```

docs/installation-process.md
```diff
@@ -0,0 +1,40 @@
...
+
+Later when the version of a package dependency needs to be switched, Packit just changes the symlink in the `dependencies` directory to point at the new dependency version.
+
+## Future additions
+There are two things that we will try to add in future Packit versions. The installation process is not yet parallalized, doing this would probably result in a significant performance increase. The installation of a single package is also not yet atomic. Meaning that if a package installation is cancelled half way through, Packit will be in a half state. Most of the time it possible to fix this with the [`fix`](./commands/fix.md) command, because it is quite advanced. However, it would be better if this was not the case, so we will try to make package installations atomic in future verions.
```
BraveChicken1:
```suggestion
There are two things that we will try to add in future Packit versions. The installation process is not yet parallelized, doing this would probably result in a significant performance increase. The installation of a single package is also not yet atomic. Meaning that if a package installation is cancelled half way through, Packit will be in a half state. Most of the time it possible to fix this with the [`fix`](./commands/fix.md) command, because it is quite advanced. However, it would be better if this was not the case, so we will try to make package installations atomic in future versions.
```

BraveChicken1 (changes requested):

docs/build-env.md
Max13245: Mhmm, alright, I'll leave this as is then. For now.

BraveChicken1 (approved):

BraveChicken1 (approved):

PR #177, opened 2026-09-03
Add notice fields to the metadata

Mrpanda-Developer:
This feature adds a optional "install_message" before installing the package. In case of certain quirks with certain packages.

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -52,7 +52,8 @@ See the tables below for all different fields, look at [Target fields](#target-f
 | `use_<script>`                  | Needs to be set to true when the script should be used. (Only for `preinstall`, `postinstall` and `uninstall`) |
 | `skip_symlinking`               | When set to true, the package is not symlinked after installation, preventing the package to be detectable through the PATH. |
 | `revisions`                     | A list of strings containing a description of what changed in each metadata or script revision. |
-| `comments`                      | A list of messages shown before installing this package version.               |
+| `preinstall_notice`             | A message shown immediately before the preinstall step. A target-specific notice overrides this value. |
+| `postinstall_notice`            | A message shown immediately after the postinstall step. A target-specific notice overrides this value. |
 | `deprecation`                   | Defines when the version deprecates, disables and the reason.                  |
 | `script_args`                   | A table of key-value pairs containing arguments passed to scripts.             |
 | `external_test_files`           | A list of external test files that are needed for executing the test script. These files are automatically downloaded. |
@@ -138,6 +139,8 @@ Targets are specified as `[targets.<bounds>]`, where bounds specify the supporte
 | `skip_symlinking`               | When set to true, the package is not symlinked after installation, preventing the package to be detectable through the PATH. Overrides the value defined in the global field. |
 | `<script-type>_script`          | Defines the name of the script to use instead of the default script name.            |
 | `use_<script>`                  | Overwrites the global `use_<script>` field. (Only for `preinstall`, `postinstall` and `uninstall`) |
+| `preinstall_notice`             | Overrides the global preinstall notice for the target.                               |
+| `postinstall_notice`            | Overrides the global postinstall notice for the target.                              |
 | `script_args`                   | A table of key-value pairs containing arguments passed to scripts, additional to the args defined in the global field. |
 | `source`                        | Defines which source to use, required when multiple sources are defined.             |
 | `external_test_files`           | A list of external test files that are needed for executing the test script for this target, additional to the files specified in the global field. These files are automatically downloaded. |
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -52,6 +52,7 @@ See the tables below for all different fields, look at [Target fields](#target-f
 | `use_<script>`                  | Needs to be set to true when the script should be used. (Only for `preinstall`, `postinstall` and `uninstall`) |
 | `skip_symlinking`               | When set to true, the package is not symlinked after installation, preventing the package to be detectable through the PATH. |
 | `revisions`                     | A list of strings containing a description of what changed in each metadata or script revision. |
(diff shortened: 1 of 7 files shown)


BraveChicken1:
I think this change is a good idea, but I'm not sure if we should have two different fields in the `package.toml` file and `targets.toml` file. I think a message is mostly version specific, so having the field only in the `targets.toml` file is probably enough.

There are also two other things that would probably be nice to add:
- This field is now only for messages before an install, it would probably be nice to have two options: `preinstall_notice` and `postinstall_notice`.
- The notices probably need to be different on different platforms, so adding the field to both `PackageVersionMeta` and `PackageTarget` would give us the option to specify it globally for a version, or only for a specific target.

One final remark, these messages should probably be printed in the `Installer::install_package` function instead, next to the calls to `execute_preinstall` and `execute_postinstall`.

Max13245:
I'm not sure if that many fields for comments is necessary, it might be. We could specify in the comment what it is meant for. Maybe it's nice to first only implement it more generally on the install. If we notice that we have a lot comments and need them to be more specific we can always add this functionality later. I do think it's good to already differentiate between targets though, because they are often very different.

BraveChicken1:
My idea was to replace the two fields that are added in this PR, by a `preinstall_notice` and `postinstall_notice` field, to differentiate between the moment the message should be shown to the user.
For `ca-certificates` the `postinstall_notice` could be useful: it could show the user that the `cert.pem` file will be generated from the certificates that are installed in the OS.

Max13245:
Ahh, so with `pre` and `postinstall` you just mean before and after the installation? I thought you meant you wanted to add something like this for all scripts (pre and postinstall being examples of script names), which seemed a bit overkill :)

BraveChicken1:
Yes, exactly. I think a separation like that is useful, and the field name explains what it does clearly.

Mrpanda-Developer:
Hey, those suggestions sound good to me, will implement them soon!

Mrpanda-Developer:
I looked at your suggestions and implemented them, could you please give me some feedback on  the changes wether they are good or not?

docs/metadata.md
```diff
@@ -54,6 +54,8 @@ See the tables below for all different fields, see [Target fields](#target-field
 | `use_<script>`                  | Needs to be set to true when the script should be used. (Only for `preinstall`, `postinstall` and `uninstall`) |
 | `skip_symlinking`               | When set to yes, the package is not symlinked after installation, preventing the package to be detectable through the PATH. |
 | `revisions`                     | A list of strings containing a description of what changed in each metadata or script revision. |
+| `preinstall_notice`             | A message shown immediately before the preinstall step. A target-specific notice overrides this value. |
+| `postinstall_notice`            | A message shown immediately after the postinstall step. A target-specific notice overrides this value. |
```
BraveChicken1: I think it would be better to use additive behaviour here: if a notice is present on both the version and the target, both notices are shown after each other.

src/installer/installer.rs
```diff
@@ -234,8 +239,12 @@ impl<'a> Installer<'a> {
 
         self.execute_postinstall(&package_id, install_meta, &install_directory, &script_args)?;
 
+        if let Some(notice) = target_meta.postinstall_notice.as_ref().or(version_meta.postinstall_notice.as_ref()) {
+            println!("{notice}");
+        }
+
         // Get the target information from the package version info
-        let target = version_meta.get_target(&install_meta.target_bounds)?;
+        let target = target_meta;
```
BraveChicken1: Please remove this line, and replace the use of `target` in the rest of the function with `target_meta`.

BraveChicken1 (changes requested):

docs/metadata.md
BraveChicken1:
```suggestion
| `preinstall_notice`             | A message shown immediately before the preinstall step.                                         |
| `postinstall_notice`            | A message shown immediately after the postinstall step.                                         |
```

src/installer/installer.rs
```diff
@@ -191,6 +191,11 @@ impl<'a> Installer<'a> {
 
         let version_meta = &install_meta.version_metadata;
         let script_args = version_meta.get_script_args(&install_meta.target_bounds)?;
+        let target_meta = version_meta.get_target(&install_meta.target_bounds)?;
+
+        if let Some(notice) = target_meta.preinstall_notice.as_ref().or(version_meta.preinstall_notice.as_ref()) {
+            println!("{notice}");
+        }
```
```suggestion
        if let Some(notice) = &version_meta.preinstall_notice {
            println!("{notice}");
        }
        if let Some(notice) = &target_meta.preinstall_notice {
            println!("{notice}");
        }
```

src/installer/installer.rs
```diff
@@ -234,14 +239,15 @@ impl<'a> Installer<'a> {
 
         self.execute_postinstall(&package_id, install_meta, &install_directory, &script_args)?;
 
-        // Get the target information from the package version info
-        let target = version_meta.get_target(&install_meta.target_bounds)?;
+        if let Some(notice) = target_meta.postinstall_notice.as_ref().or(version_meta.postinstall_notice.as_ref()) {
+            println!("{notice}");
+        }
```
```suggestion
        if let Some(notice) = &version_meta.postinstall_notice {
            println!("{notice}");
        }
        if let Some(notice) = &target_meta.postinstall_notice {
            println!("{notice}");
        }
```

src/installer/installer.rs
```diff
@@ -234,14 +239,15 @@ impl<'a> Installer<'a> {
...
 
         self.determine_active(install_meta, &package_id, target)?;
```
```suggestion
        self.determine_active(install_meta, &package_id, target_meta)?;
```

BraveChicken1 (changes requested):
You forgot a few changes, see the suggestions below.

Mrpanda-Developer:
whoops, yeah just saw. Should be fixed now.

BraveChicken1 (approved):

PR #178, opened 2026-09-03
Add local metadata

BraveChicken1:
This PR implements a local metadata storage to prevent needing network access to load in some information about packages.

Things to still decide on:
- Do we want to keep it in `<package>/.packit`? An alternative could be a global directory in the Packit prefix, also preparing for future package specific local metadata.
- How do we want to handle the packaging and the prebuilds.toml file? I don't think we want to store the prebuilds.toml file, as it is not something that a user normally needs. Maybe we should just require repository connection for packaging?
- Do we want to put the `conflicts_with` field in the local metadata, or keep it on the package in the register? Moving it would facilitate a future expansion to conflicts between specific versions, but does require some refactors.
- How do we want to handle the `homepage` and `description` fields of a `package.toml` file? Should we keep them in the register, or do we also want to include them into the local metadata?


Max13245:
- Yes, I think it would be nice to put `.packit` into something like `<prefix>/local/<package-name>`. Then we separate the actual package data from Packit data.
- Maybe they could just create a prebuild for their own target. Disregarding the prebuilds.toml content. So if we find a prebuild 'match' for a specific target it always works. Alternatively we could only store the correct `group` of the current target.
- Why would it not be possible to implement version specific conflicts in the register? The `conflicts_with` field is difficult, because it is metadata, but it's also `installation data`. I think it might be more practical/nicer to put it in the local metadata, because then we only compare repo metadata and local metadata (so it would be only metadata, not sometimes metadata other times installation/register data). I'm really not sure though.
- I think it's nice to put all repository metadata (which is also currently stored in the register) in the local metadata. Then there is a clear division between the two.

BraveChicken1:
Thank you! I would then suggest implementing the following things:
- The local metadata will get stored in `<prefix>/local/<package-name>/<package-version>/<metafile>`. Although I'm not fully sure if `local` is the best choice for the directory name, since `/usr/local` is a default directory on Unix systems with a different meaning.
- We store the prebuild metadata that is applicable for the current target in the `metadata.toml` file too.

For `conflicts_with`, `homepage` and `description`, we could add a package local metadata file, however this raises new questions:
- Do we store this local package metadata in `<prefix>/local/metadata.toml`?
- Do we also refresh this package metadata when a package version is refreshed?
- How do we handle multiple package versions from different repositories here? Their metadata could be different.

We could leave the package local metadata for now, and first try to solve the multiple package versions from different repositories in the current structure.
The `conflicts_with` could be stored on the package version, instead of the package. But this means we have to refactor the conflict detection code. Another question regarding this is whether we should take action when a new conflicting package is added after refreshing the metadata. Should this trigger an unlink of the current package or conflicting package?

src/cli/commands/init.rs
```diff
@@ -150,5 +145,35 @@ impl HandleCommand for InitArgs {
...
             1,
         );
+
+        // Get the installed package from the register
+        let Some(installed_package_version) = register.get_package_version_mut(&package_id) else {
+            error!(msg: "Packit cannot be initialized: newly created register does not contain the Packit version");
+            exit(1);
+        };
+
+        // Create the repository provider to fetch Packit metadata from
```
Max13245:
```suggestion
        // Create the repository provider to fetch Packit metadata
```

src/register/metadata/mod.rs
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pub mod error;
+mod handler;
+
+pub use handler::LocalMetaHandler;
+pub use handler::LocalMetadata;
+pub use handler::LocalPrebuildMetadata;
```
Max13245, line 11: These don't seem to be used?

src/register/package_register.rs
```diff
@@ -126,6 +126,8 @@ impl PackageRegister {
             dependents: HashSet::new(),
             install_path: install_path.into(),
             revisions: package_version.revisions.clone(),
+            last_metadata_refresh: DateTime::default(), // Initialize to UNIX epoch
+            last_metadata_change: DateTime::default(),  // Initialize to UNIX epoch
```
Max13245: I'm not sure, but wouldn't we want this inside of the local metadata? I can also understand that this is not really metadata, so that's why it's here?

src/repositories/types/prebuilds.rs
```diff
@@ -95,4 +95,20 @@ impl PrebuildsList {
...
+            addition: None,
+            version_intervals: VersionIntervals::default(),
+        };
+
+        let prebuild_meta = PrebuildMeta {
+            targets: vec![target_bounds],
+            exclude_paths: Vec::new(),
+        };
+
+        (target.architecture.to_string(), prebuild_meta)
```
Max13245: Why do we return `target.architecture.to_string()` while `target` is a parameter to the function?

src/cli/commands/update.rs
```diff
@@ -46,6 +46,10 @@ pub struct UpdateArgs {
     /// Exclude packages when using the `--all` flag, specified with <PACKAGE-NAME> ...
     #[arg(long, requires = "all")]
     exclude: Vec<PackageName>,
+
+    /// Only refresh the local metadata
+    #[arg(long)]
+    refresh_only: bool,
```
Max13245, line 52: Maybe also add a `skip_refresh` flag.

src/integrity/verifier/package.rs
```diff
@@ -69,6 +73,7 @@ fn check_package_alterations(package_id: &PackageId, register: &PackageRegister,
         disable_prebuilds: false,
     };
 
+    // TODO: prebuild list and prebuild provider are needed here
```
Max13245: There is still a TODO here.

src/register/metadata/handler.rs
```diff
@@ -0,0 +1,362 @@
...
+
+    /// Checks if a package has conflicts with any package that is installed.
+    /// Returns a list of the names of all conflicting packages.
+    pub fn get_conflicting_packages(
+        &self,
+        register: &PackageRegister,
+        package_name: &PackageName,
+        package_conflicts: &[PackageName],
+    ) -> Result<Vec<PackageName>> {
+        let mut conflicting_packages = Vec::new();
```
Max13245: This could be a hashset.

src/register/metadata/handler.rs
```diff
@@ -0,0 +1,362 @@
...
+        Ok(conflicts)
+    }
+
+    /// Checks if a package has conflicts with any package that is installed.
+    /// Returns a list of the names of all conflicting packages.
+    pub fn get_conflicting_packages(
+        &self,
+        register: &PackageRegister,
+        package_name: &PackageName,
+        package_conflicts: &[PackageName],
```
Max13245: This could also be a hashset (with some extra changes).

src/register/metadata/handler.rs
```diff
@@ -0,0 +1,362 @@
...
+            let uninstall_script_path = package_version_meta.get_uninstall_script_path(&target_bounds)?;
+            let uninstall_script_destination = metadata_dir.join(format!("uninstall.{SCRIPT_EXTENSION}"));
+            let new_file = self.request_file(provider, &uninstall_script_path, true)?;
+            if self.write_file_if_changed(&before_files, &mut after_files, uninstall_script_destination, new_file)? {
+                updated = true;
+            }
+        }
+
+        // Remove files that are not needed anymore
+        let removed_files: Vec<_> = before_files.iter().filter(|x| !after_files.contains(x)).collect();
```
Max13245: I think you can remove the collect here.

Max13245 (changes requested):
nice :)

Will you also update the docs with local metadata explanations?

Please don't forget to add rust docs and the clippy changes.

docs/commands/update.md
```diff
@@ -34,3 +34,9 @@ The `--all` flag can be used to update all latest installed versions to the late
 
 ### `--exclude <PACKAGE-NAME> ...`
 The `--exclude` flag can be used to exclude certain packages when using the `--all` flag.
+
+### `--refresh-only`
+The `--refresh-only` flag can be used to only refresh the local metadata of the specified packages. It cannot be used together with the `--skip-refresh` flag.
+
+### `--skip-refresh`
+The `--skip-refresh` flag can be used to skip the refreshing of the local metadata of the packages. It cannot be used together with the `--refresh-only` flag.
```
Max13245:
```suggestion
The `--skip-refresh` flag can be used to skip the refresh of the local metadata of the packages. It cannot be used together with the `--refresh-only` flag.
```

docs/README.md
```diff
@@ -4,11 +4,14 @@ The docs folder contains all documentation of Packit.
...
+| ------------------------------------------------- | -------------------------------------------------------------------------- |
+| [Repository metadata](./metadata.md)              | Describes the structure of metadata repositories and all available fields. |
+| [Metadata checks](./meta-checks.md)               | Describes how the metadata checks work.                                    |
+| [Tests](./tests.md)                               | Describes how Packit's tests work and how to run them.                     |
+| [Structure and Configuration](./structure.md)     | Describes Packit's file structure and how to configure Packit.             |
+| [Verifier](./verifier.md)                         | Describes how to use the verifier and repairer, and its structure in the code. |
+| [Commands](./commands)                            | Contains explanations and examples of all Packit commands.                 |
+| [Build Environment](./build-env.md)               | Describes the environment that is used for building packages.              |
+| [Installation Process](./installation-process.md) | Describes the full process that Packit executes when installing a package. |
```
Max13245:
```suggestion
| [Installation Process](./installation-process.md) | Describes the Packit installation and build process. |
```

docs/README.md
```diff
@@ -4,11 +4,14 @@ The docs folder contains all documentation of Packit.
...
+| [Register](./register.md)                         | Explains the use of the `Register.toml` file and `metadata` directory.     |
```
Max13245: Otherwise it sounds like people should be using these files:
```suggestion
| [Register](./register.md)                         | Explains the `Register.toml` file and `metadata` directory.     |
```

docs/register.md
```diff
@@ -0,0 +1,15 @@
...
+
+This file stores package information, for example the active version and the symlinked state.
+
+It also stores package version information, which consists of installation path, dependencies, source metadata repository and other fields that describe the package installation.
+
+# Local Metadata
+The local metadata is stored in the `metadata` directory. This storage consists of the relevant package metadata that is required for all operations on packages that require the metadata. By storing this data locally, we do not rely on a remote repository. If the device is offline or the repository is not available anymore, the packages keep working as before.
+
+This storage is structured in the same way as the `packages` directory, it contains directories for each package, with directories for each version of the package inside. Inside these version directories, the metadata for that version is stored. Scripts and other files that are needed are downloaded to this directory. The `metadata.toml` file stores all relevant metadata fields of the package.
```
Max13245: Maybe also explain how this helps with remote metadata changes and the local package installation being out of sync otherwise.

docs/structure.md
```diff
@@ -5,6 +5,7 @@ This file explains the Packit structure and configuration.
 ## File Structure
 The Packit directory, which contains Packit data is called the [prefix](#prefix) directory. This directory contains the following files and directories:
 - [Register.toml](#registertoml)
+- [metadata](#local-metadata)
```
Max13245, line 8: The other fields point to a section in the file, but don't mind changing it. I'm going to clear this section up anyway, because at the moment it's a bit unclear.

src/installer/installer.rs
Max13245: This is not an issue from this PR, but we should only check this if we were planning to symlink. I installed zlib, then zlib-ng-compat, got the warning "cannot symlink, because of conflicts". Then removed zlib, tried to symlink, which isn't possible according to the metadata of zlib-ng-compat.
```suggestion
        if should_symlink {
        let local_meta_handler = LocalMetaHandler::new(&self.config.prefix_directory);
        let conflicts =
            local_meta_handler.get_conflicting_packages(self.register, &package_id.name, &install_meta.package_metadata.conflicts_with)?;
if !conflicts.is_empty() {
warning!("Skipping symlinking because of conflicting packages:");
standard_print::print_list(conflicts.iter().map_styled());
should_symlink = false;
}
}
```

src/register/metadata/handler.rs
```diff
@@ -0,0 +1,368 @@
...
+    #[serde(default, skip_serializing_if = "HashSet::is_empty")]
+    #[serde(serialize_with = "serialization::serialize_set_sorted")]
+    pub external_test_files: HashSet<String>,
+
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    #[serde(serialize_with = "serialization::serialize_map_sorted")]
+    pub script_args: HashMap<String, String>,
+
+    pub deprecation: Option<DeprecationInfo>,
+    pub skip_symlinking: bool,
```
Max13245, line 58: Maybe add a default value and a `skip_serializing_if` here as well, like in the `PackageVersionMeta`.

src/cli/commands/link.rs
```diff
@@ -96,57 +95,27 @@ impl HandleCommand for LinkArgs {
...
+        let local_meta_handler = LocalMetaHandler::new(&config.prefix_directory).get_package(package_id);
+        let local_metadata = local_meta_handler.read_metadata().unwrap_or_exit_msg("Unable to read local metadata", 1);
 
-        // Skip if the package version target metadata defines skip_symlinking
-        if let Some(true) = target.skip_symlinking {
+        // Skip if the local metadata defines skip_symlinking
+        if local_metadata.skip_symlinking {
             warning!("The package metadata defines we should not symlink this package");
             return false;
         }
```
Max13245: Same issue as earlier, the `skip_symlinking` field should be checked before the conflicts.

src/integrity/verifier/package.rs
```diff
@@ -627,30 +609,18 @@ fn check_package_test(package_id: &PackageId, register: &PackageRegister, config
...
+    let local_metadata = local_meta_handler.read_metadata()?;
 
-    let provider = match provider::create_metadata_provider(&repository) {
-        Some(provider) => provider,
-        None => {
-            warning!("Cannot create metadata provider for '{package_id}', skipping check");
-            return Ok(false);
-        },
+    // Copy test script to tempfile if it exists
+    let script_text = match local_meta_handler.read_file(&format!("test.{SCRIPT_EXTENSION}")) {
```
Max13245: Maybe create a local_meta_handler::test_script_name()? This could also avoid confusion. It can be confusing that `test` can always be used, because it is simplified in the local metadata.

src/installer/installer.rs
```diff
@@ -763,57 +780,38 @@ impl<'a> Installer<'a> {
...
-            Err(e) => {
-                warning!(
-                    "Unable to read package version from metadata repository, skipping uninstall script execution. This may cause stray files"
-                );
-                warning!("{e}");
+        let local_meta_handler = LocalMetaHandler::new(&self.config.prefix_directory).get_package(package_id);
+        let local_metadata = local_meta_handler.read_metadata()?;
+
+        // Copy uninstall script to tempfile if it exists
+        let script_text = match local_meta_handler.read_file(&format!("uninstall.{SCRIPT_EXTENSION}")) {
```
Max13245: Same comment here for the use of &format!("uninstall.{SCRIPT_EXTENSION}").

src/integrity/verifier/package.rs
```diff
@@ -627,30 +609,18 @@ fn check_package_test(package_id: &PackageId, register: &PackageRegister, config
...
-    let target = version_meta.get_target(&target_bounds)?;
-
-    // Get the test script data
-    let script_path = version_meta.get_test_script_path(&target_bounds)?;
-    let script_args = version_meta.get_script_args(&target_bounds)?;
-    let script_text = provider.read_file(&package_id.name, &script_path)?;
-    let Some(script_text) = script_text else { return Ok(false) };
     let downloaded_script = scripts::write_script_to_tempfile(&script_text)?;
+
+    let script_args = local_metadata.script_args.iter().map(|(k, v)| (k.as_str(), v.as_str())).collect();
```
Max13245, line 623: Maybe we should just always use `HashMap<String, String>`, because this is done (indirectly) for almost all of these hash maps. Kind of defeats the purpose of using `HashMap<&str, &str>` I think.

src/cli/commands/update.rs
```diff
@@ -46,6 +46,14 @@ pub struct UpdateArgs {
     /// Exclude packages when using the `--all` flag, specified with <PACKAGE-NAME> ...
     #[arg(long, requires = "all")]
     exclude: Vec<PackageName>,
+
+    /// Only refresh the local metadata
+    #[arg(long, default_value = "false", conflicts_with = "skip_refresh")]
+    refresh_only: bool,
```
Max13245, line 52: The `--refresh-only` flag should also conflict with the `new_version` field, because a new version doesn't make sense when only doing a refresh.

src/cli/commands/update.rs
```diff
@@ -161,4 +182,50 @@ impl UpdateArgs {
...
+                exit(1);
+            },
+            false => &parameter_checks::expand_optional_ids(register, config, &self.packages),
+        };
+
+        for package_id in packages {
+            let Some(package_version) = register.get_package_version_mut(package_id) else {
+                error!(msg: "Expected package version {} to exist, skipping refresh", package_id.style());
+                continue;
+            };
```
Max13245, line 205: Shouldn't this just be an `.expect("Expected package version {} to exist, skipping refresh", package_id.style())`?

src/cli/commands/update.rs
```diff
@@ -161,4 +182,50 @@ impl UpdateArgs {
...
+
+            // Create repository provider for package
+            let repository = Repository::new(
+                &package_version.metadata_repository_url,
+                &package_version.metadata_repository_provider,
+            );
+            let Some(provider) = provider::create_metadata_provider(&repository) else {
+                error!(msg: "Cannot create provider for repository");
+                continue;
+            };
```
Max13245, line 215: This is an edge case and not for this PR, but still related. What do we do if a repository cannot be reached (it's just gone). Then in the register we still have that repository. The package cannot be updated, because the repository cannot be found and we don't provide an alternative. I guess this is maybe more important for the multiple repo feature.

src/register/metadata/handler.rs
```diff
@@ -0,0 +1,368 @@
...
+    /// Returns a list of conflicts defined in the metadata, which can contain packages that are not installed.
+    pub fn read_package_conflicts(&self, package: &InstalledPackage) -> Result<HashSet<PackageName>> {
+        let mut conflicts = HashSet::new();
+
+        for version in package.versions.values() {
+            let metadata = self.get_package(&version.package_id).read_metadata()?;
+
+            for conflict in metadata.conflicts_with {
+                conflicts.insert(conflict);
+            }
```
Max13245:
```suggestion
            let metadata = self.get_package(&version.package_id).read_metadata()?;
            conflicts.extend(metadata.conflicts_with);
```

src/register/metadata/handler.rs
```diff
@@ -0,0 +1,368 @@
...
+
+        for (name, package) in register.iterate_packages() {
+            if !package.symlinked || name == package_name {
+                continue;
+            }
+
+            // Check if the package specifies this package as conflict
+            if package_conflicts.contains(name) {
+                conflicting_packages.insert(name.clone());
+            }
```
Max13245, line 133:
```suggestion
            if package_conflicts.contains(name) {
                conflicting_packages.insert(name.clone());
                continue;
            }
```

src/register/metadata/handler.rs
```diff
@@ -0,0 +1,368 @@
...
+
+        // Create metadata dir if it does not exist
+        if !metadata_dir.exists() {
+            fs::create_dir_all(&metadata_dir).err_with_path("create dirs", &metadata_dir)?;
+            updated = true;
+        }
+
+        // Collect a list of all files in the metadata directory before refreshing
+        let mut before_files = Vec::new();
+        for entry in fs::read_dir(&metadata_dir).err_with_path("read", &metadata_dir)? {
```
Max13245, line 226:
```suggestion
        // Collect a list of all files in the metadata directory before refreshing
        // Note that this assumes a flat metadata directory structure
        let mut before_files = Vec::new();
        for entry in fs::read_dir(&metadata_dir).err_with_path("read", &metadata_dir)? {
```

src/register/metadata/handler.rs
```diff
@@ -0,0 +1,368 @@
...
+        let required_packit_version = max(&package_meta.required_packit_version, &package_version_meta.required_packit_version);
+        let dependencies = package_version_meta.dependencies.iter().chain(target_meta.dependencies.iter()).cloned().collect();
+
+        // Normalize external test files to flatten directories into the name
+        let external_test_files = package_version_meta
+            .get_external_test_files(target_bounds)?
+            .iter()
+            .map(ToString::to_string)
+            .map(|x| x.replace("/", "-"))
+            .collect();
```
Max13245, line 299: Even though it's short and not yet used a lot, maybe put it in a separate function to ensure consistent behaviour in the future.

src/register/metadata/handler.rs
```diff
@@ -0,0 +1,368 @@
...
+            if new_content == old_content {
+                after_files.push(destination);
+                return Ok(false);
+            }
+        }
+
+        // Write new file data
+        fs::write(&destination, new_content).err_with_path("write", &destination)?;
+        after_files.push(destination);
+        Ok(true)
```
Max13245, line 379:
```suggestion
        // Get new content, return when the file is not available
        // True is returned when the file was present before, false if the file never existed
        let Some(new_content) = new_content else {
            return Ok(before_files.contains(&destination));
        };
        
        after_files.push(destination);

        // If the file already existed, check content equality
        if before_files.contains(&destination) {
            let old_content = fs::read(&destination).err_with_path("read", &destination)?;

            // If the file did not change, skip writing and store it in `after_files`
            if new_content == old_content {
                return Ok(false);
            }
        }

        // Write new file data
        fs::write(&destination, new_content).err_with_path("write", &destination)?;
        Ok(true)
```

src/utils/serialization.rs
```diff
@@ -0,0 +1,35 @@
...
+
+/// Serializes a `HashMap` and ensures a sorted result.
+pub fn serialize_map_sorted<S, K, V>(map: &HashMap<K, V>, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+    K: Serialize + Ord,
+    V: Serialize,
+{
+    let mut entries: Vec<_> = map.iter().collect();
+    entries.sort_by_key(|(x, _)| *x);
```
Max13245:
```suggestion
    entries.sort_by_key(|(k, _)| *k);
```

Max13245 (changes requested):
It might also be nice to add timestamps for: register changed, package changed, package version changed. Although that's maybe not for this PR.

docs/structure.md
BraveChicken1: This also points to a section in the file, I don't see the difference with the other sections here.

src/integrity/verifier/package.rs
BraveChicken1: We would then need to clone in `PackageVersionMeta::get_script_args`.

src/cli/commands/update.rs
BraveChicken1: No, because then the code would panic if this occurs. Now the code shows an error an continues with the packages it can find.

src/cli/commands/update.rs
BraveChicken1: Yes that is of course something to consider. We have these fields like they are now because we wanted to always keep a package with it's original repository. But indeed it would be neat to support some sort of migration support for this.

src/register/metadata/handler.rs
BraveChicken1: This is unfortunately not possible due to ownership of `destination`.

src/register/metadata/handler.rs
BraveChicken1: I could, but it seems that these are not needed for serde. It automatically skips serializing if the Option is `None`.

docs/register.md
BraveChicken1: What do you want me to add about this? The first paragraph already describes this a bit.

docs/structure.md
Max13245: Oops, I don't know what I was thinking here.

src/register/metadata/handler.rs
Max13245: I mean for the `skip_symlinking` field.

src/integrity/verifier/package.rs
Max13245: Mhmm, okay nvm then.

docs/register.md
Max13245: It also enables us to detect changes in metadata, which wasn't possible before. Maybe it's nice to explicitly state this?

src/cli/commands/update.rs
Max13245: Yes, but this should never happen. Shouldn't we panic or at least exit early in case something happens which shouldn't happen?

docs/register.md
BraveChicken1: What do you then exactly mean by that? Which changes are detected and what is the purpose?

src/register/metadata/handler.rs
BraveChicken1: Ah, the `skip_symlinking` field. I did not use the skip_serializing there, because it then does not rely on a default that can change. But I can of course add it if you think it's better.

src/cli/commands/update.rs
BraveChicken1: We could exit here, I wouldn't choose a panic, as that is not a good way to handle errors.

src/register/metadata/handler.rs
Max13245: I think that would be better, I don't think the default will change in this case.

docs/register.md
Max13245: Well because we have a local copy we know when the metadata is updated, the purpose being that we could warn/show users that their installation is not up-to-date. For example a new conflict is added.

BraveChicken1: That is true, however we don't do that yet. So that's probably something to add to the documentation once that's actually implemented

Max13245: sure

Max13245 (approved):

PR #179, opened 2026-09-03
Fix build path escape ability of patches and license_include

BraveChicken1:
This PR ensures the patch apply process does not allow patches to change files outside of the build directory.
It also adds a similar check for the license_include field.

I will add checks for external_test_files and the local patch paths in a separate PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 - Fix patch apply directory meta check reporting wrong issues.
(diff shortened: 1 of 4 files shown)


Max13245 (commented):
Looks good!

Maybe also add it for the external test files and other fields like the `license_include`?

BraveChicken1:
External test files is another thing, those should not go outside of the package directory in the repository. That is something to implement on the provider.

I will add it to the license_include too, as that is sort of the same.

Max13245:
Yes, but I thought you might want to implement all of these "escape dir" issues at once. If you want to do it in separate PR's that's also fine.

Max13245 (approved):

PR #77, opened 2026-04-26
Add regex search

Max13245:
This adds regex search to the search command. Making it possible to search for packages using a regex string.

PR #181, opened 2026-09-05
Fix CMAKE_PREFIX_PATH and ACLOCAL_PATH in build environment

BraveChicken1:
This PR fixes the build environment by also adding symlinked packages to the `CMAKE_PREFIX_PATH` and `ACLOCAL_PATH` environment variables.

This change probably allows us to revert https://github.com/pack-it/core/pull/72 and https://github.com/pack-it/core/pull/77 for the most parts.


Max13245 (approved):
Nice, and yes it seems we can revert the changes I made in those PR's. I tested it with zstd and it works.

PR #183, opened 2026-09-08
Refactor providers to use a wrapper

BraveChicken1:
This PR refactors the `MetadataProvider` and `PrebuildProvider` by adding a wrapper struct around them. This cleans up the code by eliminating the use of Box for the providers throughout the code.


src/integrity/verifier/package.rs
```diff
@@ -468,24 +454,18 @@ pub fn check_invalid_dependencies(packages: &Vec<PackageId>, register: &PackageR
 
 /// Checks if a given package has invalid dependencies in the register according to the repository metadata.
 /// Returns a list of invalid dependencies for the given package (can be empty).
-fn invalid_dependencies_impl(package: &InstalledPackageVersion) -> Result<Vec<(PackageId, PackageId)>> {
+fn invalid_dependencies_impl(package: &InstalledPackageVersion, config: &Config) -> Result<Vec<(PackageId, PackageId)>> {
     let package_id = &package.package_id;
     let mut invalid = Vec::new();
 
-    // Assume the package version meta can be found (otherwise no issue is returned)
```
Max13245, line 475: This comment is for all of these changes. Before, if the package couldn't be found we did an early return. The argument being, a package that cannot be found can't have any issues. Now we do return the error. This is not necessarily bad, but we would need to have local metadata checks so that the verifier won't return errors.

Maybe for now do an early return, with the comment that we should implement local metadata checks.

Max13245 (changes requested):

Max13245 (approved):

PR #185, opened 2026-09-09
Extend revision functionality with local metadata

BraveChicken1:
This PR extends the support of revisions in Packit:
- Local metadata now only updates if the revisions are the same
- Revision descriptions moved from the register to the local metadata
- The init command allows defining the revision of the Packit install

Note that the change to the init command needs adjusting of the install scripts, this needs to be done with the release.

### DIFF
--- a/docs/commands/init.md
+++ b/docs/commands/init.md
@@ -1,7 +1,7 @@
 # Init
 
 The `init` command has the following command line syntax:<br>
-`pit init [--prefix <PREFIX>]`
+`pit init [--prefix <PREFIX>] [--revision <REVISION>] [--revision-descriptions <REVISION-DESCRIPTION> ...]`
 
 ## Basic init
 The `init` command initializes the Packit environment, using the following syntax:<br>
@@ -14,6 +14,12 @@ This is a complete list of all flags that can be used with the `pit init` comman
 
 ### `--prefix <PREFIX>`
 The [default prefix](../structure.md#prefix) can be overridden using the `--prefix` flag. The flag expects a directory path as its argument.
+
+### `--revision <REVISION>`
+The revision of the installed Packit version that is being initialized.
+
+### `--revision-descriptions <REVISION-DESCRIPTION> ...`
+An optional list of revision descriptions to use as fallback when the local metadata cannot be constructed from remote metadata.
 
 #### Example
 To initialize Packit at /foo/bar/buz use:<br>
--- a/src/cli/commands/init.rs
+++ b/src/cli/commands/init.rs
@@ -32,7 +32,10 @@ pub struct InitArgs {
     /// The prefix to use
     #[arg(long)]
     prefix: Option<PathBuf>,
-    // TODO: get revisions as input here
+
+    /// The revision of the Packit install
+    #[arg(long)]
+    revision: Option<u64>,
 }
 
 #[cfg(unix)]
@@ -107,9 +110,11 @@ impl HandleCommand for InitArgs {
         let package_name = PackageName::from_str("packit").expect("Expected 'packit' to be a valid package name");
         let package_version = Version::from_str(packit_version!()).expect("Expected Packit version to be in the correct format");
         let package_id = PackageId::new(package_name, package_version);
+        let revision = self.revision.unwrap_or(0);
 
         let installed_package_version = InstalledPackageVersion {
             package_id: package_id.clone(),
+            revision,
             metadata_repository_provider: DEFAULT_METADATA_REPOSITORY_PROVIDER.into(),
             metadata_repository_url: DEFAULT_METADATA_REPOSITORY_URL.into(),
             prebuilds_repository_url: None,
@@ -117,7 +122,6 @@ impl HandleCommand for InitArgs {
             dependencies: HashSet::new(),
             dependents: HashSet::new(),
             install_path: packit_package_path,
-            revisions: Vec::new(),
             last_metadata_refresh: DateTime::default(), // Initialize to UNIX epoch
             last_metadata_change: DateTime::default(),  // Initialize to UNIX epoch
         };
@@ -166,7 +170,7 @@ impl HandleCommand for InitArgs {
         // Fetch Packit metadata from the default repository
         let updated = LocalMetaHandler::new(&prefix_directory)
             .get_package(&package_id)
-            .refresh(&provider, 0)
+            .refresh(&provider, revision)
             .unwrap_or_exit_msg("Packit cannot be initialized: error while retrieving Packit metadata", 1);
 
         // Update last refresh in the package version
--- a/src/cli/commands/init.rs
+++ b/src/cli/commands/init.rs
@@ -25,7 +25,10 @@ use crate::{
         metadata::{LocalMetaHandler, LocalMetaPackageHandler, LocalMetadata},
         package_register::PackageRegister,
     },
-    repositories::{metadata::MetadataProvider, types::Licenses},
+    repositories::{
+        metadata::MetadataProvider,
+        types::{LicenseIdentifier, Licenses},
+    },
     utils::{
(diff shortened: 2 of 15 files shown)


docs/commands/init.md
```diff
@@ -15,6 +15,12 @@ This is a complete list of all flags that can be used with the `pit init` comman
 ### `--prefix <PREFIX>`
 The [default prefix](../structure.md#prefix) can be overridden using the `--prefix` flag. The flag expects a directory path as its argument.
 
+### `--revision <REVISION>`
+The revision of the installed Packit version that is being initialized.
```
Max13245:
```suggestion
The revision number of the installed Packit version that is being initialized.
```

src/cli/commands/init.rs
```diff
@@ -32,6 +42,14 @@ pub struct InitArgs {
...
     #[arg(long)]
     prefix: Option<PathBuf>,
+
+    /// The revision of the Packit install
+    #[arg(long)]
+    revision: Option<u64>,
+
+    /// The descriptions of the revisions to use as fallback
+    #[arg(long, num_args = 1..)]
+    revision_descriptions: Vec<String>,
```
Max13245, line 52: Is it maybe nice to make the `revision_descriptions` conflict with the `revisions` field? Then the revision number wil just be equal to the number of given descriptions.

src/cli/commands/init.rs
Max13245, line 160: Now the register is saved twice.

src/cli/commands/update.rs
```diff
@@ -231,7 +234,14 @@ impl UpdateArgs {
 
             // Refresh metadata
             let local_meta = LocalMetaHandler::new(&config.prefix_directory).get_package(package_id);
-            let updated_metadata = local_meta.refresh(&provider).unwrap_or_exit_msg(&format!("Cannot refresh metadata of {package_id}"), 1);
+            let updated_metadata = match local_meta.refresh(&provider, package_version.revision) {
+                Ok(updated_metadata) => updated_metadata,
+                Err(LocalMetadataError::MetadataRevisionMismatch) => continue, // Skip refresh if the revisions do not match
```
Max13245: Maybe place a debug print here if the refresh is skipped. I would say specifically not a warning, because we are trying to "hide" the metadata refresh a bit from the average user.

src/register/metadata/handler.rs
```diff
@@ -277,6 +285,25 @@ impl<'a> LocalMetaPackageHandler<'a> {
...
+
+        let metadata_file = metadata_dir.join(METADATA_FILENAME);
+        let local_meta_str = toml::ser::to_string(&metadata)?;
+
+        // Write metadata to file
+        fs::write(&metadata_file, local_meta_str).err_with_path("write", &metadata_file)?;
+
+        Ok(())
+    }
+
```
Max13245: Instead of creating the function like this, where it could potentially be misused. It might be nicer to have function that write the default Packit metadata. Then you could even check if the current local metadata handler is for Packit and return an error if it's not.

src/cli/commands/init.rs
```diff
@@ -152,28 +177,73 @@ impl HandleCommand for InitArgs {
...
+            dependencies: Vec::new(),
+            test_requirements: Vec::new(),
+            external_test_files: HashSet::new(),
+            script_args: HashMap::new(),
+            deprecation: None,
+            skip_symlinking: false,
+            conflicts_with: HashSet::new(),
+            revisions,
+            prebuild: None,
+        };
```
Max13245, line 245: If we do it like this the license identifier field doesn't have to be public for this single use case:
```suggestion
       let license = LicenseIdentifier::from_str("GPL-3.0-only").unwrap_or_exit(1);
        let local_meta = LocalMetadata {
            required_packit_version: None,
            license: Licenses::Single(license),
            dependencies: Vec::new(),
            test_requirements: Vec::new(),
            external_test_files: HashSet::new(),
            script_args: HashMap::new(),
            deprecation: None,
            skip_symlinking: false,
            conflicts_with: HashSet::new(),
            revisions,
            prebuild: None,
        };
```

Max13245 (changes requested):
Are you planning to leave the 'revision update' for another PR?

Also, don't forget to update the new provider implementations here after merging the new provider wrapper (or the other way around).

src/cli/commands/init.rs
BraveChicken1: Yes, that is done intentionally. By saving twice, we ensure there is no invalid state if the metadata cannot be stored. A future local metadata check in the verifier would then be able to fix it.

src/cli/commands/init.rs
BraveChicken1: That's possible, but isn't this safer? I don't know if the `num_args` works that well in all cases.

src/register/metadata/handler.rs
BraveChicken1: I added the comment because of that. I can of course do that, but that would mean the initialize behaviour is spread throughout the code. My idea here was that we should try to keep the initialize behaviour in the init command as much as possible.

BraveChicken1:
Yes, I'll leave the revision update for a later PR, since that requires a good design. Updating revisions requires builds to be atomic, which is currently not the case.

src/cli/commands/init.rs
Max13245: I don't think this is safer, I think it's more issue prone, because you can easily do it wrong by mismatching the parameters. As far as I have seen the `num_args` seem to work.

src/register/metadata/handler.rs
Max13245: That's a nice idea, but the current implementation also spreads the initialisation behaviour, because this function is/should only be used for Packit initialisation. I think the added check would make it safer and less bug prone.

src/cli/commands/init.rs
Max13245: Mhmm, yes. I see we do this in multiple places. Of course the ideal solution would be to save te register just before we return the error, but this annoying to implement. Maybe something we should look at in the future then.

src/cli/commands/init.rs
BraveChicken1: Yes the num_args does seem to work, but if the user puts in the wrong number of descriptions, or forgets to escape spaces, the number of revisions is incorrect, resulting in a very hard to fix state. The implementation already checks if the number of revisions and descriptions match, so mismatching is not an issue here.

src/register/metadata/handler.rs
BraveChicken1: I can change the function name to `write_packit_metadata` and add a check in it, then the local metadata struct creating is still in the init command. What do you think of that?

Max13245: Yes that would also be fine!

src/cli/commands/init.rs
Max13245: I think you misunderstood me? I mean that if the user only passes the descriptions, the number of descriptions determines the number of revisions.

BraveChicken1: I don't know if you saw the first check in the handle command, but it checks if the number of `revision_descriptions` matches the `revision`. The `revision_descriptions` are used as a fallback when the repository cannot be used to read the actual descriptions. The `revision` flag always determines the number of revisions, the descriptions are just there in case the local metadata cannot be downloaded.

Max13245: Yes, I saw we check if the number revisions is equal to the number of descriptions. But if the number of descriptions determines the number of revisions this would not be necessary. I'm not saying this is wrong, I'm saying that there might be a cleaner solution.

BraveChicken1: I understand your point, having two fields does not seem like the nicest solution here. Although I think having a separate number field is nice, to give just this little bit of guidance to the user. It prevents that the revision will get the wrong value when someone forgets to escape spaces in the description.

Max13245 (approved):

PR #186, opened 2026-09-09
Improve CI compatibility

BraveChicken1:
This PR improves the CI compatibility of Packit by addressing a few issues:
- All commands now exit with a correct exit code
- The update command now has a `--yes` flag. The install command also still needs this, but requires a variety of other flags too.
- A new `PACKIT_DISABLE_PROMPTS` environment variable ensures prompting results in an error, ensuring CI does not hang on prompts.

### DIFF
--- a/src/cli/commands/search.rs
+++ b/src/cli/commands/search.rs
@@ -206,7 +206,7 @@ impl SearchArgs {
         let latest_version = match manager.read_latest_supported_version(&repository_id, &package, &Target::current()) {
             Ok(version) => version,
             Err(RepositoryError::PackageNotFoundError { reason, .. }) => {
-                println!("Package cannot be found: {reason}");
+                error!(msg: "Package cannot be found: {reason}");
                 exit(1);
             },
             Err(e) => {
--- a/src/cli/commands/search.rs
+++ b/src/cli/commands/search.rs
@@ -198,7 +198,7 @@ impl SearchArgs {
             Err(RepositoryError::PackageNotFoundError { reason, .. }) => not_found::repository_package(package_name, manager, reason),
             Err(e) => {
                 error!(e, "Cannot read package");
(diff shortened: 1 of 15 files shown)


src/cli/commands/search.rs
```diff
@@ -207,11 +207,11 @@ impl SearchArgs {
             Ok(version) => version,
             Err(RepositoryError::PackageNotFoundError { reason, .. }) => {
                 println!("Package cannot be found: {reason}");
-                return;
+                exit(1);
```
Max13245, line 210: This should be `error!` instead of `println!`.

src/cli/display/prompts.rs
```diff
@@ -77,6 +85,11 @@ pub fn ask_user_input(question: &str) -> Result<Option<String>> {
 
 /// Prompts the user to press enter to continue.
 pub fn wait_for_continue() {
+    if prompts_disabled() {
+        // Panic for now, to prevent unnecessary usage of `Result`
+        panic!("Cannot wait for continue, since prompts are disabled");
+    }
```
Max13245, line 90: Please use the `DisplayError::UserPromptsDisabled` like in the other functions.

src/integrity/metadata/check.rs
```diff
@@ -655,12 +656,25 @@ impl MetaCheck {
             println!("{issue}");
 
             if matches!(issue.issue_type, IssueType::Fatal) {
-                return;
+                return Some(IssueType::Fatal);
```
Max13245: I know I wrote this, but can you remove this early return, because it's unnecessary :)

Max13245 (changes requested):
Looks good, only issue is that there are other places where prompts could be skipped.

Max13245 (approved):

PR #187, opened 2026-09-10
Add new unit tests

Max13245:
This PR adds new unit tests where possible. It also fixes the following issues/bugs:
- The `is_cyclic` from the tree now also checks cycles to the root.
- The `calculate_priority` function now also correctly calculates the priority for os or architectures with an addition and version.
- The parsing for dependencies and targets bounds doesn't allow for `name@` anymore. When '@' is used a version or version interval is expected, if it doesn't exist there will be an error.
- There is now parsing for target additions, which doesn't allow for empty additions or additions with invalid characters.
- Small fixes in the `License` double recursion is now properly handled and a recursive field with one item is now displayed correctly.
- The `include_license` field is now not allowed to reference symlinks anymore.


src/installer/types/dependency.rs
```diff
@@ -18,16 +28,13 @@ impl<'de> Deserialize<'de> for Dependency {
         D: serde::Deserializer<'de>,
     {
         let string: String = de::Deserialize::deserialize(deserializer)?;
-        let index = string.chars().position(|c| c == '@');
 
-        let (name, version) = match index {
-            Some(index) => string.split_at(index),
+        let (name, version) = match string.split_once('@') {
+            Some(value) if value.1.is_empty() => return Err(serde::de::Error::custom(DependencyError::ExpectedVersion)),
```
BraveChicken1:
```suggestion
            Some(value) if value.1.is_empty() => return Err(D::Error::custom(DependencyError::ExpectedVersion)),
```

src/repositories/types/license.rs
```diff
@@ -79,3 +90,108 @@ impl Display for Licenses {
...
+    fn format_single() {
+        let license = Licenses::Single("MIT".to_string());
+        assert_eq!(license.to_string(), "MIT".to_string());
+    }
+
+    #[test]
+    fn format_single_with_single_exception() {
+        let license = Licenses::SingleWithExceptions {
+            name: "MIT".to_string(),
+            exceptions: vec!["Test Exception".to_string()],
```
BraveChicken1: I would use `Test-Exception` here, as SPDX identifiers do not allow spaces.

src/repositories/types/license.rs
```diff
@@ -79,3 +90,108 @@ impl Display for Licenses {
...
+        assert_eq!(recursive_license.to_string(), "Test");
+
+        // Double recursive
+        let recursive_license = Licenses::Any {
+            any: vec![recursive_license, license],
+        };
+        assert_eq!(recursive_license.to_string(), "Test OR Test");
+
+        let license = Licenses::Any { any: Vec::new() };
+        assert_eq!(license.to_string(), "");
```
BraveChicken1: This is the same as the test case below.

src/repositories/types/license.rs
```diff
@@ -79,3 +90,108 @@ impl Display for Licenses {
...
+    }
+
+    #[test]
+    fn format_empty_recursive() {
```
BraveChicken1, line 268: Shouldn't these result in `Unknown`?

src/repositories/types/package.rs
```diff
@@ -59,3 +59,89 @@ impl PackageMeta {
...
+        },
+        platforms::{OsVersion, TargetArchitecture},
+    };
+
+    use super::*;
+
+    /// A helper method to create a simple package meta test structure
+    fn create_package_meta(versions: Vec<Version>, supported_versions: HashMap<TargetBounds, VersionIntervals>) -> PackageMeta {
+        PackageMeta {
+            name: create_package_name("test"),
```
BraveChicken1: The register uses a similar function, but allows setting the package name.

src/utils/env.rs
```diff
@@ -49,4 +49,85 @@ impl Environment {
...
+        let mut env = Environment::new();
+        env.insert_var("test", "some_value");
+        env.insert_var("test2", "");
+        assert_eq!(
+            env.env_vars,
+            HashMap::from([
+                ("test".to_string(), "some_value".to_string()),
+                ("test2".to_string(), "".to_string())
+            ])
+        );
```
BraveChicken1, line 82: I would also check if env.stripped_vars is empty here

src/utils/env.rs
```diff
@@ -49,4 +49,85 @@ impl Environment {
...
+        assert_eq!(env.env_vars, vars);
+        assert_eq!(env.stripped_vars, HashSet::new());
+    }
+
+    #[test]
+    fn insert_vars_empty() {
+        let mut env = Environment::new();
+        env.insert_vars(HashMap::<String, String>::new());
+        assert_eq!(env.env_vars, Environment::new().env_vars);
+        assert_eq!(env.stripped_vars, Environment::new().stripped_vars);
```
BraveChicken1: I would check if they are empty instead.

src/utils/env.rs
```diff
@@ -49,4 +49,85 @@ impl Environment {
...
+        env.insert_var("test3", "test_value3");
+        assert_eq!(
+            env.env_vars,
+            HashMap::from([
+                ("test".to_string(), "test_value".to_string()),
+                ("test3".to_string(), "test_value3".to_string())
+            ])
+        );
+        assert_eq!(env.stripped_vars, HashSet::from(["test2".to_string()]));
+    }
```
BraveChicken1, line 142: Maybe also add a test for the expand function.

src/utils/tree.rs
```diff
@@ -256,3 +261,107 @@ impl<V> DisplayNode<V> for () {
...
+
+    #[test]
+    fn cyclic_tree() {
+        let package_id = create_package_id("test@1");
+        let root = Node::new(package_id.clone(), (), ());
+        let mut tree = Tree::new(root);
+
+        let dependency = create_package_id("dependency@2");
+        let dependency_node = Node::new(dependency.clone(), (), ());
+        _ = tree.add_node(0, dependency_node);
```
```suggestion
        tree.add_node(0, dependency_node).unwrap();
```

src/utils/tree.rs
```diff
@@ -256,3 +261,107 @@ impl<V> DisplayNode<V> for () {
...
+
+        let other = create_package_id("other@3");
+        let other_node = Node::new(other.clone(), (), ());
+        _ = tree.add_node(1, other_node);
```
```suggestion
        tree.add_node(1, other_node).unwrap();
```

src/utils/tree.rs
```diff
@@ -256,3 +261,107 @@ impl<V> DisplayNode<V> for () {
...
+    fn filter_children() {
+        let package_id = create_package_id("test@1");
+        let root = Node::new(package_id.clone(), (), ());
+        let mut tree = Tree::new(root);
+
+        assert_eq!(tree.get_children_ids_filtered(tree.get_root(), |_| true), HashSet::new());
+
+        let dependency = create_package_id("dependency@2");
+        let dependency_node = Node::new(dependency.clone(), (), ());
+        _ = tree.add_node(0, dependency_node);
```
```suggestion
        tree.add_node(0, dependency_node).unwrap();
```

src/utils/tree.rs
```diff
@@ -256,3 +261,107 @@ impl<V> DisplayNode<V> for () {
...
+
+        assert_eq!(
+            tree.get_children_ids_filtered(tree.get_root(), |_| true),
+            HashSet::from([dependency.clone()])
+        );
```
BraveChicken1, line 337: Shouldn't we also check `|_| false`, or maybe another check? This does not fully verify that the filter is working.

src/utils/tree.rs
```diff
@@ -256,3 +261,107 @@ impl<V> DisplayNode<V> for () {
...
+    }
+
+    #[test]
+    fn format() {
+        let package_id = create_package_id("test@1");
+        let root = Node::new(package_id.clone(), (), ());
+        let mut tree = Tree::new(root);
+        let dependency = create_package_id("dependency@2");
+        let dependency_node = Node::new(dependency.clone(), (), ());
+        _ = tree.add_node(0, dependency_node);
```
```suggestion
        tree.add_node(0, dependency_node).unwrap();
```

src/utils/tree.rs
```diff
@@ -256,3 +261,107 @@ impl<V> DisplayNode<V> for () {
...
+
+        let other = create_package_id("other@3");
+        let other_node = Node::new(other.clone(), (), ());
+        _ = tree.add_node(0, other_node);
```
```suggestion
        tree.add_node(0, other_node).unwrap();
```

src/utils/tree.rs
```diff
@@ -256,3 +261,107 @@ impl<V> DisplayNode<V> for () {
...
+
+        let foo = create_package_id("foo@2");
+        let foo_node = Node::new(foo.clone(), (), ());
+        _ = tree.add_node(1, foo_node);
```
```suggestion
        tree.add_node(1, foo_node).unwrap();
```

src/utils/patches.rs
```diff
@@ -317,3 +317,47 @@ fn contains_git_prefix(operation: &FileOperation<[u8]>) -> bool {
...
+        assert!(destination.exists());
+    }
+
+    #[test]
+    fn resolve_paths_test() {
+        let source = tempdir().unwrap();
+        let source = source.path().join("test.txt");
+        let destination = tempdir().unwrap();
+        let destination = destination.path();
+        let test_file = destination.join("test.txt");
```
BraveChicken1:
```suggestion
        let destination = destination.path().join("test.txt");
```

src/utils/patches.rs
```diff
@@ -317,3 +317,47 @@ fn contains_git_prefix(operation: &FileOperation<[u8]>) -> bool {
...
+        assert_eq!(resolve_paths(&source, &test_file), (test_file.as_path(), test_file.as_path()));
+    }
+
+    #[test]
+    fn source_not_resolved() {
+        let source = tempdir().unwrap();
+        let source = source.path().join("foo.txt");
+        let destination = tempdir().unwrap();
+        let destination = destination.path();
+        let test_file = destination.join("test.txt");
```
BraveChicken1:
```suggestion
        let destination = destination.path().join("test.txt");
```

src/utils/patches.rs
```diff
@@ -317,3 +317,47 @@ fn contains_git_prefix(operation: &FileOperation<[u8]>) -> bool {
...
+        File::create(&test_file).unwrap();
+        assert_eq!(resolve_paths(&source, &test_file), (source.as_path(), test_file.as_path()));
+        assert!(!source.exists());
```
BraveChicken1, line 359:
```suggestion
        assert!(!source.exists());
        assert!(destination.exists());
```

src/cli/parameter_checks.rs
```diff
@@ -88,3 +88,103 @@ pub fn contains_package_id(packages: &Vec<OptionalPackageId>, package_id: &Packa
...
+
+        let packages = vec![create_optional_id("test@3.4.1")];
+        let duplicates = get_duplicates(&packages);
+        assert_eq!(duplicates, HashSet::new());
+
+        let mut packages = vec![create_optional_id("test@3.4.1"), create_optional_id("test@3.4.2")];
+        let duplicates = get_duplicates(&packages);
+        assert_eq!(duplicates, HashSet::new());
+
+        packages.push(create_optional_id("test"));
```
BraveChicken1:
```suggestion
        let mut packages = vec![create_optional_id("test@3.4.1"), create_optional_id("test@3.4.2"), create_optional_id("test")];
```

src/repositories/types/target_bounds.rs
```diff
@@ -10,7 +11,11 @@ use crate::{
     platforms::{Os, OsVersion, Target, TargetArchitecture},
 };
 
+static VALID_ADDITION_NAME: &str = r"^[a-z0-9\-_]+$";
```
BraveChicken1:
```suggestion
const VALID_ADDITION_NAME: &str = r"^[a-z0-9\-_]+$";
```

src/repositories/types/target_bounds.rs
```diff
@@ -92,11 +103,37 @@ impl TargetName {
...
+        if !ADDITION_NAME_REGEX.is_match(string) {
+            return Err(TargetBoundsError::InvalidAdditionName);
+        }
+
+        Ok(Self(string.to_string()))
+    }
+}
+
+impl Display for TargetAddition {
+    /// Formats an `TargetAddition` into the following format: <name>.
```
BraveChicken1:
```suggestion
    /// Formats a `TargetAddition` into the following format: <name>.
```

src/repositories/types/target_bounds.rs
```diff
@@ -92,11 +103,37 @@ impl TargetName {
...
 }
 
+/// Represents the name of an addition.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct TargetAddition(String);
+
+impl FromStr for TargetAddition {
+    type Err = TargetBoundsError;
+
+    /// Parses a string into an `TargetAddition`.
```
BraveChicken1:
```suggestion
    /// Parses a string into a `TargetAddition`.
```

src/repositories/types/target_bounds.rs
BraveChicken1: If we deny uppercase in target additions, we should probably normalize the distro name in `platforms::os`.

src/repositories/types/target_bounds.rs
```diff
@@ -269,3 +324,184 @@ impl TargetBounds {
...
+        };
+
+        let target_bounds = TargetBounds {
+            name: TargetName::Architecture(TargetArchitecture::Unknown(None)),
+            addition: None,
+            version_intervals: create_version_intervals(""),
+        };
+
+        assert!(!target_bounds.satisfied_by(&target))
+    }
```
BraveChicken1, line 464: We should probably also test `TargetName::Os(Os::Unknown)`.

src/utils/io.rs
```diff
@@ -70,7 +70,7 @@ pub fn create_folder_symlinks(original_dir: &Path, link_dir: &Path, overwrite: b
 
 /// Searches for symlinks with a certain destination (destinations inside of the destination are also a match) and removes them.
 pub fn remove_symlinks(search_dir: &Path, destination_dir: &Path) -> symlink::Result<()> {
-    if !search_dir.exists() {
+    if !search_dir.exists() || !destination_dir.exists() {
```
BraveChicken1: This check should not be here, since it means broken symlinks can never be removed using this function.

src/utils/io.rs
```diff
@@ -161,3 +161,200 @@ pub fn get_last_extension(file_name: &str) -> Option<&str> {
...
+
+        assert!(create_folder_symlinks(original.path(), &missing_destination, false).is_ok());
+        assert!(symlink::exists(&missing_destination.join("test.txt")));
+    }
+
+    #[test]
+    fn create_folder_symlinks_missing_original() {
+        let original = PathBuf::from_str("/does/not/exist").unwrap();
+        assert!(create_folder_symlinks(&original, &PathBuf::from_str("/foo").unwrap(), false).is_ok());
+        assert!(!original.exists());
```
BraveChicken1: This should probably use a temp directory

src/utils/io.rs
```diff
@@ -161,3 +161,200 @@ pub fn get_last_extension(file_name: &str) -> Option<&str> {
...
+        );
+
+        assert_eq!(
+            normalize_path(&PathBuf::from_str("foo/../../../..").unwrap()),
+            PathBuf::from_str("../../..").unwrap()
+        );
+    }
+
+    #[test]
+    fn emtpy_normalize() {
```
BraveChicken1:
```suggestion
    fn empty_normalize() {
```

src/builder/builder.rs
```diff
@@ -327,195 +286,502 @@ impl<'a> Builder<'a> {
...
+
+    use tempfile::tempdir;
 
-            fs::copy(&include_path, destination_dir.join(file_name)).err_with_path("copy", include_path)?;
+    use crate::{platforms::symlink::create_symlink, repositories::types::FileSize};
+
+    use super::*;
+
+    /// A helper script which creates a test source struct.
+    fn create_source(exclude: HashSet<String>, include: HashSet<String>) -> Source {
```
BraveChicken1:
```suggestion
    fn create_source(license_exclude: HashSet<String>, license_include: HashSet<String>) -> Source {
```

src/builder/builder.rs
```diff
@@ -327,195 +286,502 @@ impl<'a> Builder<'a> {
...
+        let build_dir = tempdir().unwrap();
+        let build_dir = build_dir.path();
+        let destination = tempdir().unwrap();
+        let destination = destination.path();
+        File::create(&build_dir.join("License")).unwrap();
+        File::create(&build_dir.join("COPYING")).unwrap();
+
+        let source = &create_source(HashSet::from(["License".to_string()]), HashSet::new());
+        copy_license_files(build_dir, destination, source).unwrap();
+        assert!(!destination.join("license").exists());
```
BraveChicken1:
```suggestion
        assert!(!destination.join("License").exists());
```

src/builder/builder.rs
```diff
@@ -327,195 +286,502 @@ impl<'a> Builder<'a> {
...
+        let build_dir = tempdir().unwrap();
+        let build_dir = build_dir.path();
+        let destination = tempdir().unwrap();
+        let destination = destination.path();
+        File::create(&build_dir.join("License")).unwrap();
+        File::create(&build_dir.join("COPYING")).unwrap();
+
+        let source = &create_source(HashSet::from(["*".to_string()]), HashSet::new());
+        copy_license_files(build_dir, destination, source).unwrap();
+        assert!(!destination.join("license").exists());
```
BraveChicken1:
```suggestion
        assert!(!destination.join("License").exists());
```

BraveChicken1 (changes requested):
Looks really good!

src/repositories/types/license.rs
Max13245: Isn't it then nicer to throw an error if the license is empty when it shouldn't be? Otherwise this is weird indirect behaviour.

src/utils/env.rs
Max13245: The reason I did it like this is that if there ever is some kind of default default value, this test would not break. This way we just test if nothing breaks, not just if they're still empty.

src/repositories/types/license.rs
BraveChicken1: Returning an error from the `Display` is not possible I think. The `from_str` could indeed return an error for this kind of behaviour.

src/utils/env.rs
BraveChicken1: I think that would then require changing the test either way. I think it's neat to check for expected behaviour (empty) here.

src/repositories/types/license.rs
Max13245: Alright I'll implement an error on the `from_str` then.

src/utils/env.rs
Max13245: The test is about if the env is unchanged, not if the vars are empty.

src/installer/types/dependency.rs
```diff
@@ -18,16 +31,13 @@ impl<'de> Deserialize<'de> for Dependency {
         D: serde::Deserializer<'de>,
     {
         let string: String = de::Deserialize::deserialize(deserializer)?;
-        let index = string.chars().position(|c| c == '@');
 
-        let (name, version) = match index {
-            Some(index) => string.split_at(index),
+        let (name, version) = match string.split_once('@') {
+            Some(value) if value.1.is_empty() => return Err(Error::custom(DependencyError::ExpectedVersion)),
```
BraveChicken1:
```suggestion
            Some(value) if value.1.is_empty() => return Err(D::Error::custom(DependencyError::ExpectedVersion)),
```

src/utils/io.rs
```diff
@@ -161,3 +161,201 @@ pub fn get_last_extension(file_name: &str) -> Option<&str> {
...
+
+        assert!(create_folder_symlinks(original.path(), &missing_destination, false).is_ok());
+        assert!(symlink::exists(&missing_destination.join("test.txt")));
+    }
+
+    #[test]
+    fn create_folder_symlinks_missing_original() {
+        let original = tempdir().unwrap();
+        let original = original.path().join("does").join("not").join("exist");
+        assert!(create_folder_symlinks(&original, &PathBuf::from_str("/foo").unwrap(), false).is_ok());
```
BraveChicken1: Shouldn't `/foo` also be in a temp directory?

src/integrity/metadata/check.rs
```diff
@@ -547,27 +547,18 @@ impl MetaCheck {
...
+            let description = format!("List of license exceptions from {} is empty", package_id.style());
             self.issues.push(MetaIssue::default(description));
         }
 
-        // Check exceptions if specified
-        if let Some(exceptions) = exceptions {
-            // Check if the list of exceptions is empty
-            if exceptions.is_empty() {
-                let description = format!("List of license exceptions from {} is empty", package_id.style());
+        // Check if one of the excptions is an empty string
```
BraveChicken1:
```suggestion
        // Check if one of the exceptions is an empty string
```

src/repositories/types/license.rs
```diff
@@ -1,6 +1,59 @@
+use regex::Regex;
 // SPDX-License-Identifier: GPL-3.0-only
```
BraveChicken1:
```suggestion
// SPDX-License-Identifier: GPL-3.0-only
use regex::Regex;
```

src/repositories/types/license.rs
```diff
@@ -9,9 +62,9 @@ use std::fmt::Display;
...
     #[default]
     Unknown,
-    Single(String),
+    Single(LicenseIdentifier),
     SingleWithExceptions {
-        name: String,
+        name: LicenseIdentifier,
 
         #[serde(rename = "with")]
         exceptions: Vec<String>,
```
BraveChicken1:
```suggestion
        exceptions: Vec<LicenseIdentifier>,
```

src/platforms/os.rs
```diff
@@ -189,7 +190,7 @@ impl OsVersion {
         }
 
         let distro = match distro {
-            Some(distro) => distro,
+            Some(distro) => TargetAddition::from_str(&distro.to_lowercase()).ok()?,
```
BraveChicken1: This should probably also print the error, before returning `None`.

BraveChicken1 (approved):

PR #189, opened 2026-09-12
Add the config reset command

Max13245:
This PR adds the `pit config reset` command to reset the `Config.toml` to its default configuration.


src/cli/commands/config.rs
```diff
@@ -511,4 +515,21 @@ impl ConfigArgs {
...
+    fn handle_reset(&self, config: EditableConfig) {
+        println!("The following configuration will be removed:");
+        self.handle_show(config);
+        println!();
+
+        let question = "Resetting the configuration cannot be undone, are you sure you want to continue";
+        if ask_user(question, QuestionResponse::No).unwrap_or_exit(1).is_no_or_invalid() {
+            println!("Cancelled configuration reset");
+            return;
+        }
```
BraveChicken1, line 529: Maybe we could add the `--yes` flag to this command aswell? It shouldn't be that hard to implement here.

Max13245: Mhmm, I think it is better if we first look for a more general solution, because it's also missing for a bunch of other commands.

BraveChicken1: We already have it on the update command, and this is an easy place to implement it. The repository add and set-url commands already have `--unchecked`, which also skips prompts.

Max13245: Mhmm, alright, I'll implement it. We really have to look for a better/more general way of handling this though.

BraveChicken1 (approved):

PR #190, opened 2026-09-12
Add metadata caching

BraveChicken1:
This PR implements metadata caching in the `MetadataProvider`.

It uses `moka` and `once_cell` to implement a cache that handles concurrency by doing only one initialization request.


Max13245 (approved):
Really nice!

PR #194, opened 2026-09-14
Fix revision field in InstalledPackageVersion

BraveChicken1:
This PR adds `#[serde(default)]` to the `InstalledPackageVersion` from the register. This is needed to ensure backwards compatibility with the old register format.

Should the revision number maybe skip serializing if the value is 0?


Max13245:
Yes, I think it's nice if we stop serialising if the value is 0.

src/register/installed_package_version.rs
```diff
@@ -12,8 +12,8 @@ use crate::{config::Repository, installer::types::PackageId};
 pub struct InstalledPackageVersion {
     pub package_id: PackageId,
 
-    // The default on `revision` is required to ensure backwards compatibility
-    #[serde(default)]
+    #[serde(default = "default_revision")]
+    #[serde(skip_serializing_if = "is_default_revision")]
     pub revision: u64,
```
Max13245, line 17: I think this is also possible, is there a reason you opted for a separate function?
```
#[serde(default, skip_serializing_if = "is_default_revision")]
pub revision: u64,
```

BraveChicken1: The `is_default_revision` function is required in both cases. I chose to also use a function for the default to ensure we use the same default in both cases.

Max13245 (approved):

PR #195, opened 2026-09-16
Add toml repairer

Max13245:
This PR adds a Toml repairer which recovers valid parts of broken Toml for the `Config.toml` and `Register.toml`.


src/integrity/repairer/initial.rs
```diff
@@ -3,93 +3,194 @@ use std::{
...
+    // Only set the field if it's not the default
+    if multiuser {
+        default_config.set_multiuser(multiuser);
+    }
+
+    confirm_config_construction(&default_config)
+}
+
+/// Gets the prefix from the given document. If the prefix cannot be found in the document. The function
+/// tries to get it with `get_config_prefix`. `None` is returned if both attempts fail.
```
BraveChicken1:
```suggestion
/// Gets the prefix from the given document. If the prefix cannot be found in the document it
/// tries to get it with `get_config_prefix`. `None` is returned if both attempts fail.
```

src/integrity/repairer/initial.rs
```diff
@@ -3,93 +3,194 @@ use std::{
...
+        let Some(value) = value.as_table() else { continue };
+        let Some(repository) = get_repository(value) else { continue };
+        found_repositories.insert(key.to_string(), repository);
+    }
+
+    Ok(found_repositories)
+}
+
+/// Tries to find the prefix directory. When found, the user is prompted to confirm. If it cannot be found
+/// de user is prompted to provide the prefix path. If the path cannot be found `None` is returned.
```
BraveChicken1:
```suggestion
/// the user is prompted to provide the prefix path. If the path cannot be found `None` is returned.
```

src/integrity/repairer/initial.rs
```diff
@@ -3,93 +3,194 @@ use std::{
...
             }
         }
 
         let question = "Please provide a different prefix path".to_string();
         match ask_user_input(&question)? {
             Some(path) => prefix_path = PathBuf::from(path),
 
-            // Return if no valid prefix path can be found (no possibility for reconstruction)
-            None => return confirm_config_construction(&default_config),
+            // Return None if no valid prefix path can be found (no possibility for reconstruction)
```
BraveChicken1:
```suggestion
            // Return `None` if no valid prefix path can be found (no possibility for reconstruction)
```

src/integrity/repairer/initial.rs
```diff
@@ -3,93 +3,194 @@ use std::{
...
     let register_dir = PackageRegister::get_path(prefix_path);
-    if let Ok(register) = PackageRegister::from(&register_dir) {
-        let used_repositories = get_used_repositories(&register);
-        if !used_repositories.is_empty() {
-            default_config.remove_repository(DEFAULT_METADATA_REPOSITORY_NAME);
+    let Ok(register) = PackageRegister::from(&register_dir) else {
+        println!(
+            "Could not use '{REGISTER_FILENAME}' to reconstruct repositories from '{}', using the default repositories instead",
+            prefix_path.display()
+        );
```
BraveChicken1, line 173: Should this function print this message? Shouldn't be done in the function that calls this one?

src/integrity/repairer/initial.rs
```diff
@@ -126,7 +227,26 @@ fn get_used_repositories(register: &PackageRegister) -> Vec<Repository> {
     repositories.into_iter().map(|(k, _)| k).collect()
 }
 
-/// Fixes a missing register. It considders all packages as missing and makes use of the inconsistent register fix.
+/// Tries to get the repository from a given table. If a field is optional and its value cannot be
+/// found the default is used. If a required field cannot be found `None` is returned.
+fn get_repository(table: &Table) -> Option<Repository> {
+    // Try to get all the fields, return early if the fields cannot be found and is not optional
```
BraveChicken1:
```suggestion
    // Try to get all the fields, return early if a field cannot be found and is not optional
```

src/integrity/repairer/initial.rs
```diff
@@ -138,9 +258,66 @@ pub fn fix_missing_register() -> Result<()> {
...
+    let mut packages: HashMap<PackageName, InstalledPackage> = HashMap::new();
+    for (package_name, package_item) in package_table {
+        let Ok(package_name) = PackageName::from_str(package_name) else {
+            continue;
+        };
+
+        let mut package: InstalledPackage = match toml::from_str(&package_item.to_string()) {
+            Ok(package) => package,
+            Err(_) => continue,
+        };
```
BraveChicken1, line 289: This will only collect InstalledPackages which are still fully valid right? Should we maybe do more in depth repairing in the future?

src/integrity/issue.rs
```diff
@@ -252,8 +264,11 @@ impl Issue {
     pub fn get_fix_message(&self) -> &str {
         match &self {
             Issue::IncorrectPermissions(_) => "To fix this issue we set the permissions again",
-            Issue::MissingConfig => "To fix this issue we try to reconstruct the Config.toml with data still in the Packit directory",
-            Issue::MissingRegister => "To fix this issue we try to reconstruct the Register.toml with data still in the Packit directory",
+            Issue::MissingConfig => "To fix this issue we try to reconstruct the 'Config.toml' with data still in the Packit directory",
+            Issue::MissingRegister => "To fix this issue we try to reconstruct the `Register.toml` with data still in the Packit directory",
```
BraveChicken1:
```suggestion
            Issue::MissingRegister => "To fix this issue we try to reconstruct the 'Register.toml' with data still in the Packit directory",
```

src/integrity/toml_repairer.rs
```diff
@@ -0,0 +1,180 @@
...
+        previous_index = *line_index;
+
+        block.push_str(line);
+        block.push('\n');
+        if block.parse::<DocumentMut>().is_ok() {
+            line_errors.drain(block_start..index + 1);
+            index -= index - block_start;
+            block = String::new();
+            block_start = index;
+        }
```
BraveChicken1, line 79: Could you add two more comments here to make it a bit more readable?

src/integrity/toml_repairer.rs
```diff
@@ -0,0 +1,180 @@
...
+            continue;
+        }
+
+        previous_index = *line_index;
+
+        block.push_str(line);
+        block.push('\n');
+        if block.parse::<DocumentMut>().is_ok() {
+            line_errors.drain(block_start..index + 1);
+            index -= index - block_start;
```
BraveChicken1:
```suggestion
            let block_size = index - block_start;
            index -= block_size;
```
This would prevent the clippy warning and allows you to remove the expect.

src/register/package_register.rs
```diff
@@ -35,6 +35,12 @@ pub struct PackageRegister {
 }
 
 impl PackageRegister {
+    /// Creates a new `PackageRegister` based on the given packages. Should normally never be used.
+    /// This is mainly used in the `Verifier` and init command. Under normal circumstances use `PackageRegister::from`.
+    pub fn new(packages: HashMap<PackageName, InstalledPackage>) -> Self {
+        Self { packages }
+    }
```
BraveChicken1, line 42: This function is not used in the init command.

BraveChicken1 (changes requested):

src/register/package_register.rs
Max13245: No, I think you want to use this instead of `new_empty` then? In that case I agree.

src/integrity/repairer/initial.rs
Max13245: Yes, it only collects still valid packages. I could actually check what packages are still missing and then fix those, however that's also done by a later check. So I'm not sure if we should do that here?

src/integrity/repairer/initial.rs
Max13245: Why?

BraveChicken1: I think it would nice to prevent too much prints in functions like these, the function is then a bit too tightly coupled to the function calling it.

src/integrity/repairer/initial.rs
BraveChicken1: The other check already checks that, so that should be fine. We could eventually look into repairing it more in depth by trying to read the valid fields from the package and package version and retrieving new values for the missing fields.

src/register/package_register.rs
BraveChicken1: I don't think the init command would really benefit from this. Should we then maybe remove the `and init command` from the message here?

src/integrity/repairer/initial.rs
Max13245: Yes that would definitely be possible. We could also imlepement this for the local metadata in the future.

src/integrity/repairer/initial.rs
Max13245: If I move the print statement it will be a double implementation, I could change the statement to:
```
println!("Could not use '{REGISTER_FILENAME}' to retrieve repositories from '{}', using the default repositories instead", prefix_path.display());
```

BraveChicken1: Hmmm yes that's also not ideal. The main thing here is that the function returns an empty HashMap and does not really use default repositories. If moving the println is not ideal, lets then just keep it here.

BraveChicken1 (approved):

PR #196, opened 2026-09-19
Add repository conflict checks

Max13245:
This PR adds support for multiple repositories.


src/installer/installer.rs
```diff
@@ -72,6 +72,15 @@ impl<'a> Installer<'a> {
             return Err(InstallerError::PermissionsError);
         }
 
+        // Make sure that repositories already used for installed packages which are not in the `Config.toml` anymore
+        // don't conflict with the current repository
+        let conflict_repositories = self.repository_manager.repository_conflicts_with(self.config, self.register.get_repository_names())?;
+        if !conflict_repositories.is_empty() {
+            let message: Vec<String> = conflict_repositories.iter().map(|(x, y)| format!(" - '{x}' incompatible with '{y}'\n")).collect();
+            let conflicts_msg = message.concat();
+            return Err(InstallerError::IncompatibleRepositories { conflicts_msg });
```
BraveChicken1: It would maybe be neater to put this formatting in the error, instead of here.

src/integrity/repairer/initial.rs
```diff
@@ -205,17 +101,18 @@ fn confirm_config_construction(default_config: &EditableConfig) -> Result<()> {
...
     let mut seen_repositories = HashMap::new();
     for package in register.iterate_all() {
         let repository = Repository {
-            url: package.metadata_repository_url.trim_end_matches('/').to_string(),
+            url: package.metadata_repository_url.clone(),
             provider: package.metadata_repository_provider.clone(),
             prebuilds_url: package.prebuilds_repository_url.clone(),
             prebuilds_provider: package.prebuilds_repository_provider.clone(),
             disable_prebuilds: false,
+            compatible_repositories: vec![], // TODO: Figure out compatibility based on repo's used for installed packages
```
BraveChicken1: There is still a TODO here.

src/register/package_register.rs
```diff
@@ -312,6 +308,11 @@ impl PackageRegister {
         latest
     }
 
+    /// Returns the set of all repositories names that were used for the installed packages.
+    pub fn get_repository_names(&self) -> HashSet<String> {
+        self.packages.values().flat_map(|p| p.versions.values()).map(|v| v.metadata_repository_name.clone()).collect()
```
BraveChicken1:
```suggestion
        self.iterate_all().map(|v| v.metadata_repository_name.clone()).collect()
```

src/repositories/types/repository.rs
```diff
@@ -21,4 +23,8 @@ pub struct RepositoryMeta {
     /// Specifies a suggestion of a prebuild repository to use with this metadata repository.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub prebuilds_provider: Option<String>,
+
+    /// A set of compatible repositories
+    #[serde(skip_serializing_if = "HashSet::is_empty", default)]
```
BraveChicken1, line 28:
```suggestion
    #[serde(default, skip_serializing_if = "HashSet::is_empty")]
```

If I'm correct this is the way we do it for the other fields.

src/repositories/manager.rs
```diff
@@ -57,15 +60,15 @@ impl<'a> RepositoryManager<'a> {
                 },
             };
 
+            let required_packit_version = repository_meta.required_packit_version.clone();
```
BraveChicken1: I think you don't need the clone here.

src/repositories/manager.rs
```diff
@@ -85,6 +88,38 @@ impl<'a> RepositoryManager<'a> {
             prebuild_providers.insert(id.clone(), prebuild_provider);
         }
 
+        // Check for repository conflicts
+        for (id, provider) in &metadata_providers {
+            let repository_meta = provider.read_repository_metadata().expect("Expected repository meta to exist");
+            for (inner_id, inner_provider) in &metadata_providers {
+                let inner_repository_meta = inner_provider.read_repository_metadata().expect("Expected repository meta to exist");
+                // Don't check compatibility between `id` and `inner_id`
```
BraveChicken1: What do you mean with this comment?

src/repositories/manager.rs
```diff
@@ -85,6 +88,38 @@ impl<'a> RepositoryManager<'a> {
...
+                if repository_meta.name == inner_repository_meta.name
+                    || repository_meta.compatible_repositories.contains(&inner_repository_meta.name)
+                    || inner_repository_meta.compatible_repositories.contains(&repository_meta.name)
+                    || config.repositories.get(id).is_some_and(|repo| repo.compatible_repositories.contains(&inner_repository_meta.name))
+                    || config.repositories.get(inner_id).is_some_and(|repo| repo.compatible_repositories.contains(&repository_meta.name))
+                {
+                    continue;
+                }
+
+                // Repositories with `id` and `inner_id` are not compatible, add them both to the unsupported set
```
BraveChicken1:
```suggestion
                // Add both `id` and `inner_id` to the unsupported set, when they are not compatible
```

src/repositories/manager.rs
```diff
@@ -336,12 +371,40 @@ impl<'a> RepositoryManager<'a> {
         None
     }
 
+    /// Gets the given repositories that are unconfigured and conflict with configured repositories. Note that this function only checks
+    /// compatibility from the side of the configured repositories.
```
BraveChicken1:
```suggestion
    /// Gets the given repositories that are unconfigured and conflict with configured repositories.
    /// Note that this function only checks compatibility from the side of the configured repositories.
```

src/repositories/manager.rs
```diff
@@ -336,12 +371,40 @@ impl<'a> RepositoryManager<'a> {
...
+        // Get all the names which aren't in the `Config.toml` anymore
+        let repository_names = self.get_repository_names()?;
+        let filtered_names: HashSet<String> = names.into_iter().filter(|name| !repository_names.contains(name)).collect();
+        if filtered_names.is_empty() {
+            return Ok(conflicting_repositories);
+        }
+
+        // Unconfigured repositories can only be validated from one side (from the unconfigured repository side it conflicts by default)
+        // Check for all repositories in the `Config.toml` if all unconfigured repositories are listed as compatible.
+        for (id, provider) in &self.metadata_providers {
```
BraveChicken1: Does validating from one side not result in issues?
Suppose the repository.toml of repository A defines it is compatible with repository B, but not the other way around.
If we then remove A from the config, it will be considered incompatible right?

src/config.rs
```diff
@@ -67,6 +67,10 @@ pub struct Repository {
     /// True to disable prebuild usage for the repository, false otherwise
     #[serde(default)]
     pub disable_prebuilds: bool,
+
+    /// A set of compatible repositories
+    #[serde(skip_serializing_if = "Vec::is_empty", default)]
```
BraveChicken1, line 72:
```suggestion
    #[serde(default, skip_serializing_if = "Vec::is_empty")]
```

CHANGELOG.md
```diff
@@ -17,17 +17,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Parsing for target additions, which doesn't allow for empty additions or additions with invalid characters.
 - License identifiers are now checked for invalid characters and being empty.
 - The `PACKIT_DISABLE_PROMPTS` environment variable, which triggers an error when Packit tries to prompt.
-- The `LocalMetadataExistence` verifier check.
-- The `config reset` command, to reset the Packit configuration file to its default.
-- Caching of the metadata from all requests to metadata repositories.
-- The verifier can now recover parts of syntactically wrong toml files, specifically the `Config.toml` and `Register.toml`.
+- Support for multiple repositories and the `compatible_repositories` field in the `Config.toml`.
```
```suggestion
- Support for multiple repositories conflict checking, using the `compatible_repositories` field in the `Config.toml` and `repository.toml`.
```

BraveChicken1 (changes requested):

src/repositories/types/repository.rs
Max13245: Well, we could in the future, but the license field above it also has it like this.

src/repositories/manager.rs
Max13245: Ah, I meant: "Don't check compatibility between `id` and `inner_id` with equal repository names". I'll adjust it.

src/repositories/manager.rs
Max13245: No, this is actually more restrictive. Repositories that are not in the config anymore cannot specify compatibility with other repositories, because they cannot be reached anymore (no URL in config). We could get this from the installed packages, however, the user has already chosen not to use/trust that repository anymore, so I don't think we should use this. Compatibility can now only be specified by the used/trusted repositories, making it more save.

src/cli/commands/config.rs
```diff
@@ -313,8 +313,11 @@ impl ConfigArgs {
 
         // Only run checks if unchecked is not enabled
         if !unchecked {
+            let config = Config::from(&Config::get_default_path()).unwrap_or_exit_msg("Cannot load config", 1);
+            let manager = RepositoryManager::new(&config);
```
BraveChicken1: The manager creation could be moved to the check_metadata_repository_compatibility function.

src/repositories/manager.rs
```diff
@@ -399,6 +399,22 @@ impl<'a> RepositoryManager<'a> {
         Ok(conflicting_repositories)
     }
 
+    /// Checks if a new repository would give conflicts. Returns true for conflicts, false otherwise.
```
BraveChicken1: It returns an option, not a boolean.

BraveChicken1 (changes requested):

src/integrity/repairer/initial.rs
Max13245: Yes, I wasn't sure about this. My solution here would be to first get all the used repositories, and because we know that they were all compatible we could list them as such in the config. Meaning each repository would list every other repository as compatible. This is not so nice though. To make this prettier we could get the repository metadata and check if the repositories were compatible anyway. 

This is a bit messy though, especially the last part.

BraveChicken1: Yeah that does indeed sound a bit messy, but I don't think there is a better way to do this.

src/repositories/types/repository.rs
BraveChicken1: Ah yeah I see, I'll note it down for refactoring.

src/repositories/manager.rs
BraveChicken1: That is a good argument, let's keep it like this then!

BraveChicken1 (approved):

PR #197, opened 2026-09-19
Add packit update exception to RepositoryManager

BraveChicken1:
This PR adds an exception for requesting Packit metadata in the RepositoryManager to allow updating from older versions.


src/repositories/manager.rs
```diff
@@ -597,8 +614,15 @@ impl<'a> RepositoryManager<'a> {
     /// A helper method to get the metadata provider.
     /// Returns a `RepositoryNotFoundError` if no repository with the given `repository_id` can be found.
     pub fn get_metadata_provider(&self, repository_id: &str) -> Result<&MetadataProvider> {
+        self.get_metadata_provider_skip_supported(repository_id, false)
+    }
+
+    /// A helper method to get the metadata provider.
+    /// If `skip_supported` is true, the repository supported check if skipped.
```
```suggestion
    /// If `skip_supported` is true, the repository supported check is skipped.
```

Max13245 (approved):

PR #198, opened 2026-09-19
Add getting started docs

BraveChicken1:
This PR adds the getting started documentation file, to help new users getting started with Packit.


docs/tutorials/getting-started.md
```diff
@@ -0,0 +1,152 @@
...
+```
+
+Note that this requires `htop` version `3.5.0` to be installed.
+
+To remove a package from your PATH, use the `pit unlink` command, see [unlink](../commands/unlink.md).
+```sh
+pit unlink htop
+```
+
+This will result in package `htop` not being available in your PATH anymore, and thus you cannot run it anymore.
```
Max13245:
```suggestion
This will result in package `htop` not being available in your PATH anymore, therefore you cannot run it anymore.
```

docs/tutorials/getting-started.md
```diff
@@ -0,0 +1,152 @@
...
+```
+
+## Adding a new repository to your config
+
+Packit allows you to add multiple repositories to your configuration.
+
+This can be useful when you want to install packages that are not distributed by the Packit team itself.
+
+> [!WARNING]  
+> External repositories can contain harmful content and should therefore be trusted before adding it to the configuration of Packit.
```
Max13245:
```suggestion
> Third party repositories can contain harmful content and should therefore be trusted before adding it to the configuration of Packit.
```

Max13245 (changes requested):
Looks good, I think the order needs to be this though:
- Installing
- Uninstalling
- Updating
- Listing
- Searching
- Active and linking
- Repositories
- Checking installation

Because 99.99% of users will only care about the first 3, maybe 5 of the above topics.

Max13245 (approved):

PR #199, opened 2026-09-20
Add directory escape check

Max13245:
This PR makes sure that paths don't escape their parent directory.


src/repositories/metadata/filesystem.rs
```diff
@@ -61,25 +61,31 @@ impl MetadataProviderImpl for FileSystemMetadataProvider {
     }
 
     fn read_file_bytes(&self, package: &PackageName, file_path: &str) -> Result<Option<Bytes>> {
-        let file_path = PathBuf::from(file_path);
-        let path = self.path.join("packages").join(package.to_string()).join(file_path);
+        let parent = self.path.join("packages").join(package.to_string());
+        let complete_path = io::normalize_path(&parent.join(file_path));
+        if !complete_path.starts_with(parent) {
```
BraveChicken1: We should probably normalize self.path in the from_repository function, to ensure the prefix of the path is always normalized

src/repositories/metadata/web.rs
```diff
@@ -117,4 +103,29 @@ impl WebMetadataProvider {
 
         Ok(response.text()?)
     }
+
+    /// Requests a file from the specified package. Returns a `Response` if the request was successful.
+    /// `RepositoryError::UnsuccessfulRequest` is returned in case of failure.
+    /// If the given `file_path` escapes the parent directory or if 404 is returned `None` is returned.
```
BraveChicken1: This does not seem to be correct, it only returns `None` if the file cannot be found. In the other cases it returns an `Err`

src/utils/requests.rs
```diff
@@ -8,6 +8,27 @@ use crate::utils::packit_version::packit_version;
...
+    // Remove prefix and suffix slashes
+    let parent = parent.trim_matches('/');
+    let path = path.trim_matches('/');
+
+    let mut components = Vec::new();
+    for component in path.split('/') {
+        if component != ".." {
+            components.push(component);
+            continue;
+        }
```
BraveChicken1: Should we maybe skip `.`?

Otherwise ./../other_path would maybe still allow escaping the parent, this can also become a new test case.

src/utils/requests.rs
```diff
@@ -35,3 +56,57 @@ pub fn get<T: IntoUrl>(url: T) -> reqwest::Result<Response> {
...
+        assert!(path_escapes_dir(parent, path));
+    }
+
+    #[test]
+    fn empty_parent_paths() {
+        let parent = "";
+        let path = "/some/random/path";
+        assert!(!path_escapes_dir(parent, path));
+        let parent = "/";
+        assert!(!path_escapes_dir(parent, path));
```
BraveChicken1: What if the parent is empty, but the path starts with `..`?

BraveChicken1 (changes requested):

src/utils/requests.rs
```diff
@@ -8,6 +8,32 @@ use crate::utils::packit_version::packit_version;
...
+pub fn path_escapes_dir(parent: &str, path: &str) -> bool {
+    // Remove prefix and suffix slashes
+    let parent = parent.trim_matches('/');
+    let path = path.trim_matches('/');
+
+    let mut components = Vec::new();
+    for component in path.split('/') {
+        if component == "." && !components.is_empty() {
+            continue;
+        }
```
BraveChicken1: Why only if components is not empty?
Shouldn't leading `./` also be trimmed from the parent path?

BraveChicken1 (changes requested):

src/repositories/metadata/filesystem.rs
```diff
@@ -61,25 +61,31 @@ impl MetadataProviderImpl for FileSystemMetadataProvider {
     }
 
     fn read_file_bytes(&self, package: &PackageName, file_path: &str) -> Result<Option<Bytes>> {
-        let file_path = PathBuf::from(file_path);
-        let path = self.path.join("packages").join(package.to_string()).join(file_path);
+        let parent = self.path.join("packages").join(package.to_string());
```
BraveChicken1:
```suggestion
        let parent = self.path.join("packages").join(&package);
```

src/repositories/metadata/filesystem.rs
```diff
@@ -61,25 +61,31 @@ impl MetadataProviderImpl for FileSystemMetadataProvider {
...
         }
 
-        Ok(Some(fs::read(&path).err_with_path("read", &path)?.into()))
+        Ok(Some(fs::read(&complete_path).err_with_path("read", &complete_path)?.into()))
     }
 
     fn read_file(&self, package: &PackageName, file_path: &str) -> Result<Option<String>> {
-        let file_path = PathBuf::from(file_path);
-        let path = self.path.join("packages").join(package.to_string()).join(file_path);
+        let parent = self.path.join("packages").join(package.to_string());
```
BraveChicken1:
```suggestion
        let parent = self.path.join("packages").join(&package);
```

src/repositories/metadata/web.rs
```diff
@@ -99,9 +87,15 @@ impl WebMetadataProvider {
             return None;
         }
 
-        Some(Self {
-            url: repository.url.clone(),
-        })
+        let url = match Url::parse(&repository.url) {
+            Ok(url) => url.to_string(),
```
BraveChicken1, line 92: Do we use `parse` and `to_string` to normalize the URL?

BraveChicken1 (changes requested):

src/repositories/metadata/web.rs
Max13245: Yes, I'll put a comment there.

BraveChicken1 (approved):

